PR #6, opened 2019-03-21
Unreviewed code on master since initial commit

foolip:
This PR is for reviewing the code on master that's been added since the first commit in this project.

@KyleJu, some suggestions for how to approach this:
- Start with README.md and DESIGN.md
- Try to run the service locally
- Try to run the tests locally
- Is there unused code, or unnecessarily complex code?
- Is the code clear? Where not, ask for documentation or suggest changes
- Are there TODOs that really should be fixed?


foolip:
Looks like the Puppeteer tests are failing in CI because of missing dependencies. https://github.com/GoogleChrome/puppeteer/pull/300 had some documentation which was never merged, could try that.

foolip:
Dependencies required to fix tests are here:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md

foolip:
Alright, CI failures resolved by just not running those tests.

foolip:
@vinyldarkscratch perhaps you'd be interested in some of this code as well. I'd love to hear what you've done differently especially with IDL and what improvements you'd most like to see here.

foolip:
@Hexcles might take a look.

foolip:
@Hexcles as discussed this is low priority, but these are the files that would be most worth looking at:
- README.md and DESIGN.md
- app.js (implementation of the API to list tests and report results)
- build.js (generating the tests)

DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Running tests
+
+### Manually
+
+When pointing a browser at https://mdn-bcd-collector.appspot.com/ to run tests,
+the server keeps track of which tests to run, accepts results from each test as
+it run, and combines all of the results at the end.
+
+To start a run, the browser posts to `/api/first` and gets a first page URL to
```
Hexcles: `/api/first` and `/api/next` aren't documented?

app.js
```diff
@@ -1,36 +1,144 @@
...
+const github = require('./github')({
+  auth: `token ${secrets.github.token}`,
+});
+
+const Tests = require('./tests');
+const tests = new Tests({
+  manifest: require('./MANIFEST.json'),
+  host: process.env.NODE_ENV === 'production'
+      ? 'mdn-bcd-collector.appspot.com'
+      : 'localhost:8080',
```
Hexcles: Near the end of the file, when starting the test, `env.PORT` is checked and honoured. The same should be done here.

app.js
```diff
@@ -1,36 +1,144 @@
...
 const express = require('express');
+const cookieParser = require('cookie-parser');
+const uniqueString = require('unique-string');
+
+const logger = require('./logger');
+
+// TODO: none of this setup is pretty
+const {CloudStorage, MemoryStorage} = require('./storage');
+const storage = process.env.NODE_ENV === 'production'
+   ? new CloudStorage
```
Hexcles: So in prod, test reports are stored in GCS? Is there an expiration/lifecycle? Otherwise, is this a DoS concern?

app.js
```diff
@@ -1,36 +1,144 @@
...
+app.use(express.static('generated'));
 
-app.get('/', (req, res) => {
-  res
-    .status(200)
-    .send('Hello, world!')
-    .end();
+app.get('/api/tests', (req, res) => {
+  const {after, limit} = req.query;
+  const list = tests.list(after, limit ? +limit : 0);
```
Hexcles: Is the plus sign (`+limit`) just for coercing `limit` into int?

DESIGN.md
foolip: Oops, sent https://github.com/foolip/mdn-bcd-collector/pull/55 to refresh docs.

app.js
foolip: Good idea: https://github.com/foolip/mdn-bcd-collector/pull/56

app.js
foolip: The bucket (mdn-bcd-buffer) has a "15 days or older" retention policy. It could be much shorter, but this is enough to ensure it doesn't grow forever. It would still be possible to spam the bucket of course, but it's not possible to read it back except as pull requests on a GitHub repository, so it would be for troublemaking and not for free file hosting.

One could require authentication, but that would make it more work to test on odd browsers. For example, I ran it on a phone in a store once, and I wouldn't do that if it took any work.

app.js
foolip: Yes, this is because of the `afterIndex + 1 + limit` expression in the called function. `1 + '1'` is the string "11" in JavaScript, so I needed to coerce to a number somewhere.

PR #19, opened 2019-05-06
Add smoke test for reffy-reports integration

foolip:
This depends on the real data scraped from specs and would fail if
specs are reorganized, so the picked examples are ones which are very
unlikely to change.

### DIFF
--- a/test/integration/reffy-reports.js
+++ b/test/integration/reffy-reports.js
@@ -1,0 +1,36 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


test/integration/reffy-reports.js
```diff
@@ -0,0 +1,36 @@
+// Copyright 2019 Google LLC
```
Hexcles: Just out of curiosity: is it recommended to have the full LICENSE at the beginning of every file? I thought one top-level LICENSE file + one-line copyright header would do.

foolip: It's what was in the quickstart repo that I started this from. The short form for the existing ones I know of are with BSD licenses, so I'd have to look into how to adapt this to that form. So I didn't :)

PR #29, opened 2019-05-06
Add tests for methods in buildIDLTests

foolip:
Anticipating breakage from https://github.com/foolip/mdn-bcd-collector/pull/28.

### DIFF
--- a/test/unit/build.js
+++ b/test/unit/build.js
@@ -68,6 +68,17 @@ describe('build', () => {
       ]);
     });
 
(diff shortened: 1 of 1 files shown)


foolip:
These tests actually won't be broken by updating webidl2.js, but they do reveal that some code can be simplified now, so that's good.

foolip:
Simplifications are in https://github.com/foolip/mdn-bcd-collector/pull/28.

PR #34, opened 2019-05-07
Build tests for prefixed+unprefixed variants of CSS properties

foolip:


build.js
```diff
@@ -66,6 +66,20 @@ function collectCSSPropertiesFromReffy(reffy, propertySet) {
...
+// add prefixed forms from unprefixed and vice versa
+function expandCSSProperties(propertySet) {
+  for (const prop of propertySet) {
+    const unprefixedProp = prop.replace(/^-[^-]+-/, '');
+    if (unprefixedProp !== prop) {
+      propertySet.add(unprefixedProp);
+      // fall through to add other prefixed forms
+    }
+    for (const prefix of ['moz', 'ms', 'webkit']) {
+      propertySet.add(`-${prefix}-${unprefixedProp}`);
```
Hexcles: I'm no JavaScript expert, but modifying an iterable while iterating over it always makes me nervous. Is this safe?

foolip: Oh, I didn't do this deliberately. `s=new Set([0]); for (const i of s) { if (i < 10) { s.add(i + 1); } }; s` in node logs "Set { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }" which shows that the behavior is the reasonable one: iteration order is insertion order and things added during iteration will be iterated at the end.

https://stackoverflow.com/a/28306768 has an explanation. I'll link to that in a comment.

foolip:
@Hexcles did you mean to LGTM this % the question?

build.js
foolip: FWIW, I think I'd like to restructure these bits to just return sets instead of adding to an existing set. I did that to not have to merge sets, but that'll be just a few lines of code so I think I'll do that, later.

PR #119, opened 2019-08-22
Convert GitHub Actions into the new .yml syntax

foolip:
Done using https://github.com/actions/migrate


foolip:
The checks (triggered by push) aren't going to run, so merging this to verify it working.

PR #326, opened 2020-07-21
Introduce historical.idl for generating tests for non-standard stuff

foolip:


foolip:
@vinyldarkscratch I've pushed a bunch of changes to this repo today, and this is one you might care about. If you find that there are things in BCD not covered by mdn-bcd-collector, in many cases it should work to extend this new historical.idl file.

foolip:
@vinyldarkscratch this file is now renamed to https://github.com/foolip/mdn-bcd-collector/blob/main/non-standard.idl and I've populated it with a lot of stuff. When you compare BCD entries with what tests are generated you will no doubt discover more stuff to be added, like things from other engines (especially Gecko) and stuff that was in Chromium but already removed.

PR #372, opened 2020-08-22
Better file loading for update-bcd

queengooborg:
This PR makes a few adjustments to the file loading in the BCD update script to allow for folders as an argument, as well as warnings when it cannot parse the JSON from the files, or if somehow, the argument isn't a file or folder.


foolip (commented):
LGTM

PR #373, opened 2020-08-22
Always test constructors with no arguments

queengooborg:
A few issues had arisen in testing constructors with dummy arguments that would require some intricate parsing of the WebIDL to resolve.  This PR reworks the constructor test to always test with no arguments, and if the error says that it just doesn't have enough arguments, the constructor is present and functional.

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -100,18 +100,10 @@
           }
         } else if (subtest.property == 'constructor') {
           var iface = parentPrefix+subtest.scope;
-          var args = '';
-          for (var a = 0; a < eval(iface+'.constructor.length'); a++) {
-            if (a == 0) {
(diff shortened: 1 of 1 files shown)


static/resources/harness.js
```diff
@@ -100,28 +100,25 @@
...
             if (
               err.name == 'TypeError' &&
               err.message == 'Illegal constructor'
             ) {
               result.result = false;
-            } else if (stringIncludes(err.message, 'Failed to construct')) {
-              // If it failed to construct, and not illegal, there's a
-              // constructor
+            } else if (
+              stringIncludes(err.message, 'Not enough arguments') ||
```
foolip, line 115: Can you annotate which engine (and maybe version ranges?) uses these strings? Especially "Failed to construct" makes me curious.

PR #383, opened 2020-08-30
Update the BCD updater

vinyldarkscratch:
This PR updates the BCD update script with a few things, namely the following:

- Initialize support matrix with all browsers in BCD data -- this allows us to better determine version numbers, due to the improvement listed right below
- Improved null value handling -- if support changes between versions and there are null/unknown values in between, do not determine a version number
- Make sure there are no duplicate statements -- when adding a statement including `version_removed`, sometimes it will add a duplicate; this was resolved
- Add prefix support -- account for prefixed support in browsers

### DIFF
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -4,7 +4,6 @@ const compareVersions = require('compare-versions');
 const fs = require('fs');
 const path = require('path');
 const uaParser = require('ua-parser-js');
-const bcd = require('mdn-browser-compat-data');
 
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch thanks for hacking away at this! Just to set expectations, I normally won't be reviewing these changes even after they've been merged. Instead I'll be focused on the results coming out, which should lead back to bugs in this code which we can then think harder about.

PR #388, opened 2020-08-31
Improve version omitting in update-bcd

vinyldarkscratch:
This PR improves the version omission when there are null gaps in `update-bcd`.

### DIFF
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -234,9 +234,7 @@ function inferSupportStatements(versionMap) {
     if (supported === true) {
       if (!lastStatement) {
         statements.push({
-          version_added: (i === 0 || lastKnown.support === false) ?
-            version :
-            true,
+          version_added: (lastWasNull && lastKnown.support === false) ? true : version,
           ...(prefix && {prefix: prefix})
         });
       } else if (!lastStatement.version_added) {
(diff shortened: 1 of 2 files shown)


update-bcd.js
```diff
@@ -234,9 +234,7 @@ function inferSupportStatements(versionMap) {
     if (supported === true) {
       if (!lastStatement) {
         statements.push({
-          version_added: (i === 0 || lastKnown.support === false) ?
-            version :
-            true,
+          version_added: (lastWasNull && lastKnown.support === false) ? true : version,
```
foolip, line 237: If we have incomplete data can we instead generate a range, which a human reviewer can then try to replace? https://github.com/mdittmer/mdn-confluence does this now, and I think it's pretty useful while skimming diffs at least.

Aside: the logic for updating BCD could really be shared here. In both cases we have a set of results per browser and might have gaps, or in the case of mdn-bcd-collector, inconclusive tests which returned null.

vinyldarkscratch: By a range, do you mean like `≤79`?  I could certainly work that in, although we wouldn't be able to do that for everything, since ranged values are only allowed in BCD for certain versions (specifically, engine changes).

PR #389, opened 2020-08-31
Add custom tests for AudioBuffer[SourceNode]

vinyldarkscratch:
This PR adds a few new custom tests.

### DIFF
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -16,6 +16,15 @@
       "length": "return buffer && 'length' in buffer;",
       "numberOfChannels": "return buffer && 'numberOfChannels' in buffer;",
       "sampleRate": "return buffer && 'sampleRate' in buffer;",
+    },
+    "AudioBufferSourceNode": {
+      "__base": "var audioCtx = new (window.AudioContext || window.webkitAudioContext)(); if (!audioCtx) {return false}; var source = audioCtx.createBufferSource();",
+      "__test": "return !!source;",
+      "loop": "return source && 'loop' in source;",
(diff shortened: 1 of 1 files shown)


custom-tests.json
```diff
@@ -7,6 +7,24 @@
       "drawElementsInstancedANGLE": "return a && 'drawElementsInstancedANGLE' in a;",
       "vertexAttribDivisorANGLE": "return a && 'vertexAttribDivisorANGLE' in a;",
       "VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE": "return a && 'VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE' in a;"
+    },
+    "AudioBuffer": {
+      "__base": "var audioCtx = new (window.AudioContext || window.webkitAudioContext)(); if (!audioCtx) {return false}; var buffer = audioCtx.createBuffer(2, audioCtx.sampleRate * 3, audioCtx.sampleRate);",
+      "__test": "return !!buffer;",
+      "duration": "return buffer && 'duration' in buffer;",
```
foolip, line 14: These look very regular. It looks like the important part of this is how to create an instance of `AudioBuffer` and the rest could be generated. I think this will be a very common pattern, and that we would benefit for having a notion of factories, i.e., code that creates instances.

Some factories can be generated from interpreting Web IDL, like that `navigator.mediaDevices` is a `MediaDevices` instance. But then some should be custom factories.

No need to fix this now, just dropping some ideas.

vinyldarkscratch: Yeah, that makes sense!  I'll follow up with another PR to improve the custom test generation.

PR #396, opened 2020-09-01
Add additional non-standard interfaces

vinyldarkscratch:
This PR adds more non-standard interfaces for various interfaces starting with the letter D/E.


foolip (commented):
Good stuff! If you can sprinkle `[Exposed=Window]` on interfaces that'd be nice, but we'll catch that if we later turn on webidl2.js validation.

Issue #407, opened 2020-09-03
Include as little JS as possible prior to running tests


foolip:
This dependency was added in 8637e29b4eb3d690dfe6d07b9129f5fc0611e423 with the comment "Include core-js polyfill for Promise/Set/etc. support in older browsers".

This is in line with our use of JSON3, but adds more stuff to the global. Any generated tests for Promise/Set/etc would be invalidated by this, which is also the case for JSON3. But JSON3 has a `JSON3.noConflict()` which we could use to avoid this.

We should have as little JS as possible run before the actual tests, and once they have finished, we can pull in limitless amounts of code and stylesheets to present the results in a beautiful way :)

vinyldarkscratch:
I wonder if there's a better, smaller option for polyfilling the promises; that would help out greatly.  Unfortunately, promises are a core component in the functionality of the Worker and ServiceWorker tests -- at least, in a reasonable way.  Alas, it would certainly cause issues for automated JavaScript feature testing if we're to add that...

foolip:
Can you point to the code where we depend on promises? It's probably straightforward to replace it with something relatively antique to avoid the issue.

vinyldarkscratch:
We're using it to handle the messages back from both the Workers.  For the service workers, it's not as much of a problem since Promises were natively handled before Service Workers, but regular Web Workers were supported first.

foolip:
I see, so it's used to wait for all tests to finish. It will be a little bit of bookkeeping, but if that's all then it can be rewritten to check if all tests are finished every time one test finishes. It could also simplify things to never do any work in parallel, but I don't know if that's an important performance optimization?

foolip:
For the use of promises to await results from many worker tests, I believe https://github.com/foolip/mdn-bcd-collector/issues/436 will also help. The worker script would run all the tests and post back results in a single message, so there wouldn't be any need for the bookkeeping that `Promise.allSettled` now simplifies.

vinyldarkscratch:
I've just implemented the code to send all the tests at once, so promises are no longer needed, and thus `core-js` is no longer a dependency!

foolip:
Excellent, thanks! Let's leave this open though to see if there's more we can get rid of when things have settled down.

queengooborg:
With #1690 now merged, the only pre-test dependency is our own `harness.js`, so I think it's safe to close this issue.  I also plan to minify `harness.js` which should help even more.

Issue #376, opened 2020-08-24
Expose URLs for the individual tests


foolip:
When updating BCD based on the results, one will often want to verify using BrowserStack or Sauce Labs, especially if the proposed change seems implausible.

Currently, what I do is write a new test by hand in https://software.hixie.ch/utilities/js/live-dom-viewer/ and try that in different browsers. This is very time consuming.

I think we should have URLs for each individual test, or at least each individual BCD path. PRs on BCD should then link to these tests, so the URLs should be stable.

I have this approach in mind:
- Move test generation to the server at runtime, instead of in a build step.
- A URL like `/api/` might include all tests for api.*, while `/tests/api/HTMLMediaElement/audioTracks` would include the tests for `api.HTMLMediaElement.audioTracks`.
- When visiting one of these pages manually, the results are shown visually.
- Still have a default grouping and no visual results for automated collection.

@vinyldarkscratch WDYT?

foolip:
The "Move test generation to the server at runtime" bit could be softened to just generating a database of available tests and letting the server assemble those into responses. It's very convenient to have the full output of such a step available for diffing to see what impact a change is going to have.

vinyldarkscratch:
I like the idea of having URLs for individual tests so they can be run individually and manually, so that anyone can use the tests provided, and suggest changes if needed.  I think that a URL would make the most sense!  I think that `/tests/api/HTMLMediaElement/audioTracks` would also display the results right there, rather than send the results to the server, perhaps?

(Regarding the visualization, I was actually thinking of providing more verbosity (and add some additional UI stuff like a spinner) to tell the user that tests are happening and the page isn't frozen, for example.)

foolip:
Sounds like we're on the same page.

One minor point though. If there is any feedback while the tests are running, then it's important that the scripts involved use as archaic JavaScript as possible, and don't add any global variables or modify the DOM in any non-trivial way.

Ideally the test code runs first, so that the test code can have full control over the DOM and the scripting environment. Once completed it's fine to pull in more CSS and JavaScript to display the results. So anything that happens before the tests have run needs to be fairly minimal. Updating `document.title` or a single `<pre>Running test 3/5</pre>` would be OK I think.

foolip:
https://github.com/foolip/mdn-bcd-collector/pull/398, cool!

Reopening to file a bug report: http://mdn-bcd-collector.appspot.com/ has a dropdown, but no items in the dropdown. So I can't test if this works well or not :)

[...]

Issue #375, opened 2020-08-23
Handle conflicting results?


queengooborg:
As the results repo grows larger and larger, there's going to be a high probability of conflicts in the data for numerous reasons -- such as different exposure times in different scopes, adjustments to the tests, different implementation for different OSes, and sometimes, even patch versions of browsers (implemented in Chrome 54.0.2835 and removed in Chrome 54.0.2836).  We will need to implement conflict-processing.

What I'm thinking is the following conflict handling (in order of preference):
- Prefer higher collector version -- if the results from `mdn-bcd-collector` 0.0.5 and 0.0.6 differ, choose results from 0.0.6 (test could have been patched)
- Prefer higher patch version of browser -- if the results from Chrome 84.0.4147.89 and 84.0.4147.135 differ, choose results from 84.0.4147.135
- Prefer "true" result if different OSes -- if the results from macOS says "true" but Ubuntu Linux says "false", the result is "true"
- Prefer higher OS version -- if tests from macOS 10.14 and 10.15 have conflicting results, choose results from 10.15
- Prefer the more likely answer -- if two results files for the same version say "true" but one says "false", say "true" (in the case of 50/50 results, say "null" since we can't determine)

Of course, there will be other issues along the way -- a good example is people modifying their user agents -- but we can figure those out later.

_Issue inspired by https://github.com/mdittmer/mdn-confluence/pull/47._

foolip:
I think that sounds like the right rules. What I found in https://github.com/mdittmer/mdn-confluence/pull/47 was mostly the same browser version of different OSes or different versions of the same OS giving different results. For the latter, we should prefer the most recent version of the OS, as it might be a feature that's conditionally enabled based on OS support. For things that are different between for example macOS and Windows for the same browser, notes are the only way of representing that.

vinyldarkscratch:
Added two list entries for the OS differences!

foolip:
Hijacking this issue to remember sorting out https://github.com/foolip/mdn-bcd-collector/pull/416/files#r485522138. That's a case of conflicting results :)

PR #416, opened 2020-09-04
Server-side builds (version 0.1.0)

### LINKED ISSUE #413: Generate test database vs. HTML files
To aid in tests, we should do the following (as suggested in https://github.com/foolip/mdn-bcd-collector/issues/376#issuecomment-685021436):

- Generate a database (JSON file) of tests per BCD data path.
- Let the server use that database to generate the test pages, keeping the output of those pages as trivial as possible. The browser should not need to query and interpret MANIFEST.json to run tests.

### LINKED ISSUE #414: Move backend endpoints?
The `/api/` URL contains two separate pieces of functionality in it -- the API tests, and the API endpoints.  We should rename the API endpoint URLs to something else (for example, `/backend/` or just leave them as-is).  Either that, or move the tests themselves to a new URL, which we could do considering #413.

### LINKED ISSUE #417: Stylize pages
I was thinking that we should give the homepage, reports page, and individual test completion pages a little bit of styling to make them look more attractive.  What do you think, @foolip?

### LINKED ISSUE #418: Console API doesn't post updates correctly
In BCD, the Console API is referred to as `api.Console` -- however, the WebIDL actually defines `console` as a namespace with no capitalization.  Since the identifiers are case-sensitive, this causes no updated results.

There's a few solutions I can think of:
- Rename the identifier in the build script -- hacky and special-casing, but it'll do the trick at least
- Rename the identifier in update-bcd -- also hacky and special-casing
- Make identifiers case-insensitive -- could cause unforseen issues in BCD

### LINKED ISSUE #420: Generate endpoints on the fly in backend
> Do we need manifest.json separately from test.json, shouldn't the backend be able to derive everything it needs from just the list of tests?

_Originally posted by @foolip in https://github.com/foolip/mdn-bcd-collector/pull/416#discussion_r484376238_

Additional comments:
https://github.com/foolip/mdn-bcd-collector/pull/416#discussion_r484440344
https://github.com/foolip/mdn-bcd-collector/pull/416#discussion_r484450944

### LINKED ISSUE #437: Use <datalist> for individual test pages
As a follow-up to https://github.com/foolip/mdn-bcd-collector/issues/376, it would be nice to use an `<input list>` with `<datalist>` instead of a `<select>`, as this makes it easier to find the right endpoint by typing part of an API name.

vinyldarkscratch:
This PR provides a huge changeset that qualifies for the bump of a minor version.  Primarily, this PR re-introduces server-side builds, and performs a number of optimizations.

Updates include the following:

- Web UI
  - Add a little content to the homepage
  - Add CSS styling to the pages
  - The dropdown is now a `<datalist>`, allowing for type-to-search
  - Individual tests for APIs can be tested under a specified scope
  - There is now a dedicated 404 page
- Build/Generation
  - All tests are compiled server-side once again
  - All HTML is now generated using EJS templates
  - `MANIFESTS.json` is now reduced to just the tests, and is moved to `tests.json`
  - Endpoints are generated by the server
- Test Runtime
  - All the tests are now under `/tests/` to separate it from the backend API endpoints
  - Tests are both displayed and reported to the server in one go, with a link to submit a PR to the results GitHub repo
  - Promises are no longer used for the worker and service worker; thus, core-js is no longer a dependency
- Tests
  - `api.console` is now mapped to `api.Console`
  - The `toString` operation is now added
  - Added support custom API tests for non-members (i.e. `api.AbortSignal.abort_event` or `api.BaseAudioContext.createPeriodicWave.disableNormalisation_supported`)
  - Added test for `api.DOMTokenList.remove_whitespace_and_duplicates`
- Misc.
  - Nodemon is now used to auto-reload the server when any files change

Fixes #413, fixes #414, fixes #417, fixes #418, fixes #420, fixes #437.


vinyldarkscratch:
I'm requesting your review, @foolip, since this is a big change, and I want to make sure it's all good before we merge and deploy!

app.js
```diff
@@ -64,21 +64,70 @@ function catchError(err, res) {
...
+function generateTestPage(endpoint) {
+  const theseTests = tests.getTests(endpoint);
+  let testScope = tests.getScope(endpoint);
+  let individual = false;
+
+  const lines = [
+    '<!DOCTYPE html>',
+    '<html>',
+    '<head>',
+    '<!--Copyright 2020 Google LLC',
```
foolip, line 76: Let's drop this. A big comment at the top is annoying when viewing source to understand the test. If I hear that we need this even for generated output then we can add it back at the bottom.

foolip (commented):
This isn't my full review, I'll take a look at this next week. Very exciting change though, I think this will make things much easier to under for BCD reviewers.

app.js
vinyldarkscratch: Okay, I'll go ahead and drop it!  I think it'll be all fine, since there's copyright notices elsewhere.

vinyldarkscratch:
Whew!  I have now added prefix support into server-side builds, so we can get prefix support once again!  It functions in a similar manner to how the prefix testing worked before -- test each piece with each prefix until support is found, then move onto the next -- with different prefixed variations built into the manifest.  Unfortunately, this makes the manifest file very long and very large as a result -- I've separated it into two files.

Long story short, I think this is now ready for deploying!

static/resources/harness.js
```diff
@@ -338,7 +338,11 @@
     css.rel = 'stylesheet';
     css.type = 'text/css';
     css.href = '/resources/style.css';
-    document.head.appendChild(css);
+    try {
+      document.head.appendChild(css);
+    } catch(e) {
+      // If the CSS fails to load, oh well
```
foolip, line 344: In which browser does `document.head.appendChild(css)` throw an exception? Any exception shouldn't be due to the CSS failing to load, which happens asynchronously, but more likely because `document.head` doesn't exist. I'm guessing this happened on IE<9, based on https://developer.mozilla.org/en-US/docs/Web/API/Document/head?

build.js
```diff
@@ -728,57 +661,108 @@ function copyResources() {
         'sourceMappingURL=core.js.map'
     );
 
-    fs.writeFile(corejsPath, result, 'utf8', function(err) {
+    fs.writeFileSync(corejsPath, result, 'utf8', (err) => {
```
foolip, line 664: Putting this after the `await fs.readFile` would be good. Or, if this will all go away by not using core-js, a TODO would be fine.

build.js
```diff
@@ -728,57 +661,108 @@ function copyResources() {
...
       if (err) return console.log(err);
     });
   });
-}
+};
 
-async function build(bcd, webref) {
+const buildManifest = (tests) => {
```
foolip, line 670: Not reviewing this in detail, but I note that it's not a very long function, so the server should be able to compute all of this information from the tests listing at runtime. (See also comment in app.js about `/tests/*` for how I think this can change.)

build.js
```diff
@@ -728,57 +661,108 @@ function copyResources() {
...
-}
+
+  return manifest;
+};
+
+/* istanbul ignore next */
+const build = async (webref, bcd) => {
+  const IDLTests = buildIDL(webref);
+  const CSSTests = buildCSS(webref, bcd);
+  const tests = Object.assign(IDLTests, CSSTests);
```
foolip, line 734:
```suggestion
  const tests = Object.assign({}, IDLTests, CSSTests);
```

To avoid modifying `IDLTests` in a way that's not obvious.

build.js
```diff
@@ -728,57 +661,108 @@ function copyResources() {
...
-  await writeManifest(manifest);
-  copyResources();
-}
+
+  return manifest;
+};
+
+/* istanbul ignore next */
+const build = async (webref, bcd) => {
+  const IDLTests = buildIDL(webref);
```
foolip, line 732: By passing in `bcd` here, we should have the information we need about which prefixes to test for.

non-standard.idl
```diff
@@ -28,6 +28,34 @@ partial interface Document {
   attribute EventHandler onbeforecut;
   attribute EventHandler onbeforepaste;
   attribute EventHandler onsearch;
+
+  boolean contains();
```
foolip, line 32: Can you split out all updates to non-standard.idl into its own PR?

find-missing.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      features.push(...traverseFeatures(obj[i], identifier + i + '.'));
+    }
+  }
+
+  return features;
+};
+
+const findMissing = () => {
```
foolip, line 26: Good stuff! From that list one can see a lot that can be added to non-standard.idl. But it's best to do so separately from this overhaul.

.eslintrc.json
```diff
@@ -1,14 +1,26 @@
 {
     "extends": ["eslint:recommended", "google"],
-    "plugins": ["prettier"],
+    "plugins": ["prefer-arrow"],
```
foolip, line 3: By now it may be incredibly hard to untangle, but doing the reformatting resulting from this and other linter changes separately from the backend overhaul would be great.

What I'd do: copy over the relevant changes to package.json and .eslintrc.json changes to a new branch, and reformat using `--fix` and cherry-picking over any changes that can't be made automatically.

When that changes is on main, merge in the main branch and resolve any conflicts in favor of what's on this branch. What remains should be the real changes on this branch.

foolip (commented):
Submitting a partial review, I stopped half way through build.js.

static/resources/harness.js
vinyldarkscratch: It was actually early versions of Chrome (like Chrome 4), which led me to believe that it may occur in other browsers as well (like IE).  I tried with `document.getElementsByTagName('head')[0]` as well, but had no luck there.

.eslintrc.json
```diff
@@ -1,14 +1,26 @@
...
     "parser": "babel-eslint",
     "parserOptions": {
         "ecmaVersion": 8
     },
     "rules": {
         "comma-dangle": ["error", "never"],
-        "require-jsdoc": "off",
-        "no-unused-vars": ["error", {"varsIgnorePattern": "^_$"}]
+        "max-len": ["error", {
+            "code": 80,
```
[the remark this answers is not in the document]
vinyldarkscratch: I thought I had to redefine the entire rule if I was making any changes to it, but it looks like rule options will combine!

DESIGN.md
```diff
@@ -47,6 +52,11 @@ The following JSON...
       "__base": "var canvas = document.createElement('canvas'); var gl = canvas.getContext('webgl'); var instance = gl.getExtension('ANGLE_instanced_arrays');",
       "__test": "return canvas && instance;",
       "drawArraysInstancedANGLE": "return true && instance && 'drawArraysInstancedANGLE' in instance;"
+    },
+    "Console": {
+      "__additional": {
+        "log.substitution_strings": "console.log('Hello, %s.', 'Bob')"
```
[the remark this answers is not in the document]
vinyldarkscratch: I wanted to put some sort of filler test down in the document regardless of whether it actually functioned, just to show how the additional tests function, and then come back to replace it with a working one later.

app.js
```diff
@@ -42,27 +45,26 @@ const github = require('./github')(
 
 const Tests = require('./tests');
 const tests = new Tests({
-  manifest: require('./MANIFEST.json'),
-  host: process.env.NODE_ENV === 'production' ?
-      'mdn-bcd-collector.appspot.com' :
-      `localhost:${PORT}`,
+  tests: require('./generated/tests.json'),
+  manifest: require('./generated/manifest.json'),
```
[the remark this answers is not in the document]
vinyldarkscratch: Yeah, the list of tests should give the backend everything that it needs!  I'm thinking to save this for another time.

.eslintrc.json
vinyldarkscratch: Cherry-picked as #421!

unittest/unit/tests.js
```diff
@@ -0,0 +1,178 @@
...
+      assert.equal(
+          theseTests.next('http://host.test/tests/css/properties'),
+          null
+      );
+    });
+  });
+
+  it('listMainEndpoints', () => {
+    assert.deepEqual(tests.listMainEndpoints(), [
+      '/api/interfaces',
```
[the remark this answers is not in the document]
vinyldarkscratch: Collapsing the tests does sound nice, it ends up making the amount of tests in a single file a little too much, causing browser freezes (especially on older browsers).  I'm also not sure how to properly hook it up with our report function to submit PRs for all of the data to the results repo.  Will have to come back to this one at a later time.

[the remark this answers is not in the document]
foolip: Moved to https://github.com/foolip/mdn-bcd-collector/issues/430.

build.js
foolip: Moved to https://github.com/foolip/mdn-bcd-collector/issues/431.

.eslintrc.json
foolip: Thanks!

foolip: Done in https://github.com/foolip/mdn-bcd-collector/pull/421, thanks!

find-missing.js
foolip: Done in https://github.com/foolip/mdn-bcd-collector/pull/422.

non-standard.idl
foolip: Done in https://github.com/foolip/mdn-bcd-collector/pull/427.

build.js
foolip: Moved to https://github.com/foolip/mdn-bcd-collector/issues/431.

app.js
```diff
@@ -138,6 +141,16 @@ app.post('/api/results/export/github', (req, res) => {
       .catch(/* istanbul ignore next */ (err) => catchError(err, res));
 });
 
+app.all('/tests/*', (req, res) => {
+  const endpoint = `/${req.params['0']}`;
+
+  if (tests.listAllEndpoints().includes(endpoint)) {
```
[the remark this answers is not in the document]
foolip: Moved to https://github.com/foolip/mdn-bcd-collector/issues/420.

build.js
```diff
@@ -84,152 +65,130 @@ function getCustomTestAPI(name, member) {
...
+          prefix: prefix
+        });
       }
-      if (method === 'CSS.supports' || method === 'all') {
-        lines.push(`bcd.addTest("${ident}", ${JSON.stringify(
-            {property: name, scope: 'CSS.supports'}
-        )}, 'CSS');`);
+    }
+  } else {
+    for (const prefix1 of prefixesToTest) {
```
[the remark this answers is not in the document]
foolip: Moved to https://github.com/foolip/mdn-bcd-collector/issues/431.

DESIGN.md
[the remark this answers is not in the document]
foolip: Alright. I think `api.DOMTokenList.remove_whitespace_and_duplicates` is a good candidate for a simple custom test for illustration purposes:
```js
var elm = document.createElement('b');
elm.className = ' foo bar foo ';
elm.classList.remove('bar');
return elm.className === 'foo';
```

static/resources/harness.js
foolip: I see. Catching and ignoring the error is fine. Only the comment needs updating, something like "If inserting the link element fails" would be more accurate about where the exception is coming from. (The consequence will be the CSS won't load, though.)

DESIGN.md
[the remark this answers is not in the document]
vinyldarkscratch: Sure thing, added!

build.js
```diff
@@ -97,133 +75,114 @@ const getCustomTestAPI = (name, member) => {
   return test;
 };
 
+const getCustomSubtestsAPI = (name) => {
+  const subtests = {};
+
+  if (name in customTests.api) {
+    const testbase = customTests.api[name].__base || '';
+    if ('__additional' in customTests.api[name]) {
+      for (
```
foolip, line 84: I've filed https://github.com/foolip/mdn-bcd-collector/issues/432 to undo weird Prettier formatting like this.

static/resources/harness.js
```diff
@@ -84,182 +102,74 @@
...
-  function runCSS(callback) {
-    var results = [];
-
+  function runCSS(callback, results) {
     var length = pending.length;
     for (var i = 0; i < length; i++) {
-      updateStatus('Testing ' + pending[i].name);
-      results.push(test(pending[i]));
+      if (pending[i].scope == 'CSS') {
+        updateStatus('Testing ' + pending[i].name);
```
foolip, line 152: This isn't new, but I'm nervous about doing really anything at all while the tests are running. Let's come back to this when moving more of the code generation to the server side.

static/resources/harness.js
```diff
@@ -84,182 +102,74 @@
...
     }
 
-    pending = [];
-
     callback(results);
   }
 
-  function runWorker(callback) {
-    var results = [];
+  function runWorker(callback, results) {
```
foolip, line 172: Not in this PR, but the setup for worker tests I'm imaging when most of the of work is done on the server side is that a single large script is run in a worker, and then the results are all posted back in one go. In the window context there wouldn't be any need to keep track of pending tests.

static/resources/harness.js
```diff
@@ -360,81 +271,93 @@
...
+  function runAndDisplay() {
+    run(function(results) {
+      var css = document.createElement('link');
+      css.rel = 'stylesheet';
+      css.type = 'text/css';
+      css.href = '/resources/style.css';
+      try {
+        document.head.appendChild(css);
+      } catch (e) {
+        // If the CSS fails to load, oh well
```
foolip:
```suggestion
        // If document.head isn't supported, no CSS will be loaded. That's OK.
```

tests.js
```diff
@@ -16,45 +16,167 @@
...
 
-      if (afterIndex === -1) {
-        throw new Error(`${after} not found in test manifest`);
+    for (const [ident, test] of Object.entries(this.tests)) {
+      for (const scope of test.scope) {
+        let endpoint = '';
+        let httpsOnly = false;
+        switch (scope) {
+          case 'Window':
+            endpoint = '/api/interfaces';
```
foolip, line 34: In a follow-up, maybe `/api/?scope=Window`, modulo naming of "scope" discussed elsewhere. Not only interfaces are being tested any longer, namespaces and attributes/functions on the global scope are too.

tests.js
```diff
@@ -16,45 +16,167 @@
...
+            break;
+          case 'Worker':
```
foolip, line 36: Can both 'Worker' and 'DedicatedWorker' happen? It seems like it should be one or the other.

tests.js
```diff
@@ -16,45 +16,167 @@
...
+
+        if (endpoint) {
+          if (!(endpoint in endpoints)) {
+            endpoints[endpoint] = {
+              scope: scope,
+              httpsOnly: httpsOnly,
+              entries: []
+            };
+          }
+          if (!(ident in endpoints[endpoint].entries)) {
```
foolip, line 57: This won't work. `x in y` where `y` is an array will only be true if `x` is an array index (string or number) or if it's some Array builtin like `length` or `push`. Not the cases you want to check for, I presume.

tests.js
```diff
@@ -16,45 +16,167 @@
...
 
-      begin = afterIndex + 1;
-      if (limit) {
-        end = afterIndex + 1 + limit;
+    return endpoints;
+  }
+
+  listMainEndpoints(urlPrefix = '') {
+    return Object.keys(this.endpoints).map((item) => (
+      ['', `${urlPrefix}${item}`]
```
foolip, line 69: What does the empty string mean here? I think it's a marker for "run this test by default", is that right? Being more explicit about this would be good.

tests.js
```diff
@@ -16,45 +16,167 @@
...
+    ];
+  }
+
+  next(after) {
+    const afterURL = new URL(after);
+    if (!this.httpOnly && afterURL.protocol === 'http:') {
+      return `https://${this.host}${afterURL.pathname}`;
+    }
+    const endpoints = this.listMainEndpoints('/tests');
+    const index = endpoints.findIndex((item) => {
```
foolip, line 92: What if `index` is -1 (before we add 1)? Isn't that an error case that we should do something about? As it is, will we end up running the same test in an endless cycle?

tests.js
```diff
@@ -16,45 +16,167 @@
...
+    if (index >= endpoints.length) {
+      return null;
+    }
+
+    const endpoint = endpoints[index][1];
+
+    if (this.endpoints[endpoint.replace('/tests', '')].httpsOnly) {
+      const newUrl = `https://${this.host}${endpoint}`;
+      if (this.httpOnly) {
+        // Skip this endpoint and go to the next
```
foolip, line 105: I've filed https://github.com/foolip/mdn-bcd-collector/issues/435 to perhaps simplify this later. What I did was a bit too clever IMHO, and continuing to maintain the complexity like here is unfortunate.

tests.js
```diff
@@ -16,45 +16,167 @@
...
+        return this.next(newUrl);
       }
+      return newUrl;
+    }
+
+    return `http://${this.host}${endpoint}`;
+  }
+
+  getTests(endpoint) {
+    let idents = [];
```
foolip, line 115:
```suggestion
    let idents;
```

It's assigned in both branches, let's be explicit that this empty array is never used.

tests.js
```diff
@@ -16,45 +16,167 @@
...
+      '</head>',
+      '<body>',
+      '<p id="status">Running tests...</p>',
+      '<script>'
+    ];
+
+    for (const [ident, test] of Object.entries(theseTests)) {
+      for (const scope of test.scope) {
+        if (!testScope || scope == testScope) {
+          lines.push(`bcd.addTest("${ident}", ${JSON.stringify(test.tests)}, "${scope}");`);
```
foolip, line 171: This is a place where I expect resolving https://github.com/foolip/mdn-bcd-collector/issues/376 will change things a lot. Rather than feeding input to the client-side test generator here, we should just output stuff like:
```js
bcd.addTest('api.Document', function() {
  return 'Document' in window;
});
```

But not in this PR :)

foolip (commented):
Finally got through the whole change. Lots of comments, please enjoy :)

Let's punt as much as possible to issues where things aren't regressed by this change.

static/resources/harness.js
```diff
@@ -84,182 +102,74 @@
...
-              result.result = true;
-            } else {
-              result.result = null;
-            }
+    for (var i = 0; i < data.tests.length; i++) {
+      var test = data.tests[i];
 
-            result.message = 'threw ' + stringify(err);
+      try {
+        var value = eval(test.code);
```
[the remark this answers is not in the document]
vinyldarkscratch: The reason for `eval` is to be able to send the functions to the workers -- all messages sent to workers must be serializable, so passing in functions wouldn't work, unfortunately.

static/resources/harness.js
```diff
@@ -84,182 +102,74 @@
...
+        if (value && typeof value === 'object' && 'result' in value) {
```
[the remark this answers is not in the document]
vinyldarkscratch: This is a part of the addition of `bcd.testConstructor`, which returns an object (that can include a message).

tests.js
vinyldarkscratch: Some interfaces mark exposure for `Worker`, while others for `DedicatedWorker`.  From what I've gathered, they're both the same however.

package.json
```diff
@@ -38,11 +38,12 @@
     "json-stable-stringify": "1.0.1",
     "json3": "3.3.3",
     "mdn-browser-compat-data": "1.0.35",
+    "nodemon": "^2.0.4",
```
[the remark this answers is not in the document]
vinyldarkscratch: Overall it shouldn't interfere with GAE, but just in case, I'll split this into `start-dev` instead.  Nodemon watches for any file changes and automatically reloads the server accordingly -- SUPER helpful for development purposes.

static/index.html
```diff
@@ -1,69 +1,126 @@
...
+    function getTests() {
+      var button = document.getElementById('start');
+      var dropdown = document.getElementById('tests');
+      var individualTestOptionsBox = document.getElementById('individualTestOptionsBox');
+      var secureContext = document.getElementById('secureContext');
+      var limitScope = document.getElementById('limitScope');
 
-<p><a href="https://github.com/foolip/mdn-bcd-collector">Source on GitHub</a></p>
-</body>
+      var client = new XMLHttpRequest();
```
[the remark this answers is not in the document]
vinyldarkscratch: Since it's within the scope of server-side builds, I decided to actually tackle this, and hook up EJS to build the pages.  All pages -- including the homepage, results, and the tests themselves -- are now built with the EJS templating system!

static/resources/harness.js
vinyldarkscratch: The `updateStatus()` isn't really working due to how much is going on in the JavaScript already (the DOM isn't updating), so these can be taken out safely.

tests.js
vinyldarkscratch: I just pushed f8934be379e465cdc78a2d514d0f3a6d6c164380 which should provide some assistance for this.  In that commit, I added `?reportToServer=true`, which will tell the website to send the data to the server rather than display it.  It no longer counts on a test being a "main endpoint" to determine if it should report, which works towards eliminating main endpoints altogether.

views/index.ejs
```diff
@@ -0,0 +1,84 @@
...
+    } else {
+      limitExposureBox.style.display = "block";
+    }
+  }
+
+  button.onclick = function() {
+    var choice = getDropdownSelection(dropdown);
+    var newUrl = choice[1];
+    if (choice[0] > 0) {
+      var protocol = secureContext.checked ? 'https:' : 'http:';
```
foolip, line 71: Could this use the protocol of the current page instead? For localhost 'https:' will never make sense, so showing the option then isn't really useful.

views/layout.ejs
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <!--Copyright 2020 Google LLC
```
foolip, line 4: Let's omit copyright notices from the generated pages. I shouldn't have included it in the first place :)

views/tests.ejs
```diff
@@ -0,0 +1,38 @@
...
+          background-color: #111;
+          color: white;
+        }
+      }
+    </style>
+  </head>
+  <body>
+    <p id="status">Running tests...</p>
+    
+    <script src="/resources/json3.min.js"></script>
```
foolip, line 28: Can you add a TODO linking to https://github.com/foolip/mdn-bcd-collector/issues/407 here?

foolip (commented):
All done reviewing, phew!

I think this looks pretty good now, I'd like to get it landed so we can keep iterating in smaller PRs. I've tried deploying https://staging-dot-mdn-bcd-collector.appspot.com/ but something is broken.

Once we have a working staging deployment, it would be good to compare the collected results before/after this change, to see if there are unintended changes in there.

foolip:
It looks like the 500 error on https://staging-dot-mdn-bcd-collector.appspot.com/ is because the 256 MB limit is being exceeded. @vinyldarkscratch does that give any clues? Feel free to push changes to the staging branch to narrow down what it might be that causes this.

nodemon.json
```diff
@@ -1,4 +1,9 @@
 {
   "verbose": true,
-  "ignore": ["static/*", "coverage/*", ".nyc_output/*"],
+  "ignore": [
+    "static/*",
+    "views/*",
```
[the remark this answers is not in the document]
vinyldarkscratch: It's because the server itself doesn't need to reload for that -- just the browser.

tests.js
vinyldarkscratch: Fixed down below!  If the `index` is 0 (`-1 + 1`), we return `null` now, to prevent an endless loop.

build.js
```diff
@@ -624,12 +669,12 @@ const buildCSS = (webref, bcd) => {
...
         code: getCustomTestCSS(name) || [
-          {property: attrName, scope: 'document.body.style'},
-          {property: name, scope: 'CSS.supports'}
+          {property: attrName, owner: 'document.body.style'},
+          {property: name, owner: 'CSS.supports'}
         ],
         combinator: '||'
       },
-      scope: ['CSS']
+      exposure: ['CSS']
```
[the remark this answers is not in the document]
vinyldarkscratch: Yeah, `exposure` was certainly misused here.  I'm now adding a `category` property, which will also help future-proof for `javascript` feature testing or other categories.

static/resources/harness.js
```diff
@@ -140,16 +140,15 @@
...
 
     return result;
   }
 
   function runCSS(callback, results) {
     var length = pending.length;
     for (var i = 0; i < length; i++) {
-      if (pending[i].scope == 'CSS') {
-        updateStatus('Testing ' + pending[i].name);
+      if (pending[i].exposure == 'CSS') {
```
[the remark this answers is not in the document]
vinyldarkscratch: I've now fixed up `exposure` and removed this `runCSS()` function entirely!

build.js
```diff
@@ -624,12 +669,12 @@ const buildCSS = (webref, bcd) => {
     tests[`css.properties.${name}`] = compileTest({
       raw: {
         code: getCustomTestCSS(name) || [
-          {property: attrName, scope: 'document.body.style'},
-          {property: name, scope: 'CSS.supports'}
+          {property: attrName, owner: 'document.body.style'},
+          {property: name, owner: 'CSS.supports'}
```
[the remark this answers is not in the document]
vinyldarkscratch: `CSS.supports` is kind of a special case in this sense, as it's the only instance of "how do we test it".  We could, perhaps, call it "method" in this case?

vinyldarkscratch:
> It looks like the 500 error on https://staging-dot-mdn-bcd-collector.appspot.com/ is because the 256 MB limit is being exceeded. @vinyldarkscratch does that give any clues? Feel free to push changes to the staging branch to narrow down what it might be that causes this.

What is the 256 MB limit for?  Is it storage space, or how large the response from the server is?  (Unfortunately my knowledge of GAE is limited...)

vinyldarkscratch:
Great news: I was able to eliminate "main endpoints" in an effective manner, so we now only have endpoints for the individual tests!  My solution was to have the script both submit AND display the results, no redirects, with a link at the top to submit the results to the GitHub repo.

foolip:
@vinyldarkscratch other than this not working on AppEngine, is this ready for final review, or do you have more changes planned?

vinyldarkscratch:
I'd say this is ready for final review!

vinyldarkscratch:
I want to try to integrate the datalist if possible, and simplify the code some if I can.  I should be ready for a final review soon, though!

vinyldarkscratch:
Okay, this is now ready for final review, @foolip!

build.js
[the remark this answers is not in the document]
foolip: That would be fine, yeah. I won't block this PR on fixing this though, I've linked it in https://github.com/foolip/mdn-bcd-collector/issues/375 to not forget.

unittest/app/app.js
```diff
@@ -109,4 +106,68 @@ describe('/api/results', () => {
...
+    const res = await agent.get(`/tests/${tests[1][0].replace(/\./g, '/')}`);
+    assert.equal(res.status, 200);
+  });
+
+  it('get all tests', async () => {
+    const res = await agent.get('/tests/');
+    assert.equal(res.status, 200);
+  });
+
+  it('get a non-existent tests', async () => {
```
foolip, line 139: Good testing!

foolip (approved):
Let's get this merged!!!

Issue #10, opened 2019-03-26
Get AppEngine deployment URL at runtime


foolip:
URLs used in the manifest are hardcoded here:
https://github.com/foolip/mdn-bcd-collector/blob/9c26f56b10fcb1ca47dc92101531c2547d4aaf5e/app.js#L41

This makes it impossible to collect results from a staging instance:
https://github.com/foolip/mdn-bcd-results/pull/105

Probably https://cloud.google.com/appengine/docs/standard/nodejs/runtime#metadata_server is how to get the URL of the instance. @Hexcles do you know how wpt.fyi does this?

Hexcles:
@foolip the Go `appengine` package provides a helper function for it, which I'm pretty sure simply constructs the URL by concatenating project ID and `.appspot.com`.

foolip:
@Hexcles do you know the lines of code where this is done? If I know the name of the Go binding maybe I can find if there's an equivalent for NodeJS.

Hexcles:
@foolip see the code around here:
https://github.com/web-platform-tests/wpt.fyi/blob/8ab7e1b72bec05da62f700f85966fd15c138d4c3/shared/appengine.go#L162

PR #419, opened 2020-09-07
Deploy staging-dot-mdn-bcd-collector.appspot.com from staging branch

foolip:


foolip:
@vinyldarkscratch would this be useful for experimenting with a real deployment? Note that it's probably broken because of https://github.com/foolip/mdn-bcd-collector/issues/10 though.

vinyldarkscratch:
Having a staging version would be quite helpful, I think, yeah!

foolip:
Alright, merged. Now, whenever you'd like to test something live, just force push to the staging branch, as long as its based on f8b4fdc or more recent.

https://github.com/foolip/mdn-bcd-collector/issues/10 probably means that running through the tests doesn't work yet.

PR #433, opened 2020-09-08
Turn on ESLint's curly rule

foolip:


foolip:
@vinyldarkscratch this will conflict with other changes. It's a matter of preference too, so feel free to reject it.

vinyldarkscratch (approved):
I do enjoy this rule and use it in my own projects -- this sounds good to me!

PR #434, opened 2020-09-08
Compute the host based on GAE environment variables

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/10.


vinyldarkscratch (approved):
Looks good to me!

PR #442, opened 2020-09-10
Add custom test for MediaDevices API

vinyldarkscratch:
This PR adds a custom test for the MediaDevices API to test the instance exposed via `navigator.mediaDevices`.


custom-tests.json
```diff
@@ -41,6 +41,9 @@
     },
     "EXT_texture_filter_anisotropic": {
       "__base": "var canvas = document.createElement('canvas'); if (!canvas) {return false}; var gl = canvas.getContext('webgl'); if (!gl) {return false}; var instance = gl.getExtension('EXT_texture_filter_anisotropic');"
+    },
+    "MediaDevices": {
+      "__base": "var instance = navigator.mediaDevices;"
```
foolip, line 46: FWIW, this is something we should be able to derive from just the IDL, since it's just accessing attributes starting at the global object. But  no harm in testing it manually until then!

PR #447, opened 2020-09-11
Simplify the lookup of tests for /tests/*

foolip:
First checking if there were endpoints and then getting the tests was
doing double/unnecessary work.


vinyldarkscratch (approved):
Good catch, looks good to me!

PR #456, opened 2020-09-12
Use MessageChannel API to handle messaging with Service Worker

### LINKED ISSUE #453: Service worker tests don't function in older browsers
It seems that in older browsers, service worker tests seem not to respond -- actually, it seems that the service workers aren't installing at all for some odd reason (there's no request for `serviceworker.js`).  Will need to resolve.

Encountered in Chrome 40.

vinyldarkscratch:
This PR fixes communication with service workers due to lack of support for certain features.  By using the `MessageChannel` API, which by BCD was supported in all browsers before Service Workers were supported, we can now get messages from service workers in any browser.

Fixes #453.

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -215,11 +215,17 @@
           }).then(function(reg) {
             return window.__waitForSWState(reg, 'activated');
           }).then(navigator.serviceWorker.ready).then(function(reg) {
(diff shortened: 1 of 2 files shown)


foolip (commented):
Nice!

PR #459, opened 2020-09-12
Add instance factories for NodeIterator and TreeWalker

foolip:


custom-tests.json
```diff
@@ -49,6 +49,12 @@
     },
     "MediaDevices": {
       "__base": "var instance = navigator.mediaDevices;"
+    },
+    "NodeIterator": {
+      "__base": "var instance = document.createNodeIterator(document);"
```
foolip, line 54: @vinyldarkscratch as we keep adding these, I think it will be a very common pattern that we have a simple short expression to create the instance. What do you think about a form that takes just the expression, and the generated code would be `'detach' in (document.createNodeIterator(document))` if the expression is simple enough?

We could perhaps run Prettier on the generated code so that we don't need to worry about formatting, like whether we need the extra parenthesis.

foolip: I've deployed this to https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/TreeWalker/expandEntityReferences to illustrate. The expression there is much longer and harder to parse: `(function() {var instance = document.createTreeWalker(document);return instance && 'expandEntityReferences' in instance;})();`

vinyldarkscratch: I have some ideas on how to implement this, so I've opened up #465!

vinyldarkscratch (approved):
Looks good to me, thanks!

PR #463, opened 2020-09-12
Split non-standard.idl into files per related spec

foolip:
The generated tests.json does not change at all with this change.


foolip:
@vinyldarkscratch this isn't entirely done as non-standard.idl is not empty yet, but it's done enough to review I think.

vinyldarkscratch (approved):
Awesome, looks good to me!  👍

PR #464, opened 2020-09-12
Remove test generation for Web IDL dictionaries

foolip:
Unlike interfaces and namespaces, there are no objects corresponding to
dictionaries that can be tested.

In order to test dictionaries one has to pass an object to a method (or
constructor) that takes that dictionary as an argument, and use getters
to see which dictionary members are accessed. This is probably best done
as custom tests until a pattern emerges which can be generalized.

Bump the version to 0.2.0 since a lot of tests, albeit ones that didn't
work, are removed by this change.


vinyldarkscratch (approved):
This makes sense to me, yeah -- more often than not, dictionary objects aren't exposed on their own, so having auto-generated tests for them doesn't quite work the way we'd want it to.  We'd need to do custom tests for them.  👍

PR #475, opened 2020-09-14
Use different buckets for production and staging

foolip:
To make this work, create a new app.staging.yaml which deploys to a
different service. Versions are again left unspecified, which means that
new versions will be created for each deploy.


foolip:
I'm not certain that this will work, but will merge and see :)

PR #477, opened 2020-09-14
Handle server errors in harness.js and results pages better

vinyldarkscratch:
This PR updates the error handling a little better to prevent client crashes, by doing the following:

1) Catch 4xx errors as well as 5xx errors
2) If no parseable JSON, set `response.error` to the response's text

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -323,7 +323,7 @@
     client.send(body);
     client.onreadystatechange = function() {
       if (client.readyState == 4) {
-        if (client.status >= 500) {
+        if (client.status >= 400) {
           updateStatus('Failed to upload results.');
         } else {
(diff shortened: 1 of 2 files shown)


static/resources/harness.js
```diff
@@ -323,7 +323,7 @@
     client.send(body);
     client.onreadystatechange = function() {
       if (client.readyState == 4) {
-        if (client.status >= 500) {
+        if (client.status >= 400) {
```
foolip, line 326: @vinyldarkscratch what we should probably do is the equivalent of https://fetch.spec.whatwg.org/#ok-status, checking if that status is 2xx.

vinyldarkscratch: Sounds good, opened #482 to do so!

PR #478, opened 2020-09-14
Use form on homepage (allow "Enter" to run tests)

vinyldarkscratch:
This PR wraps the input elements in a `form` element rather than a `div`, allowing clients to use the Enter key to submit and start the tests.


foolip (commented):
Nice!

PR #485, opened 2020-09-14
Remove trailing whitespace

foolip:


foolip:
@vinyldarkscratch spotted this when diffing prod and main. Not sure if there's a lint configuration we could turn on for this?

vinyldarkscratch:
Unfortunately we wouldn't be able to use ESLint to lint for this, but Prettier we could configure to do so?  Guess this is a reason to get Prettier as well?

vinyldarkscratch:
I created an issue to discuss the beautification and linting options for this, so we can continue discussion there, and merge this PR!

PR #492, opened 2020-09-14
Don't test for "instance in self" for members

vinyldarkscratch:
This PR reverts an update that has each member test to make sure that the API is `true`, so that if the API is `false`, the members are `null`.

### DIFF
--- a/build.js
+++ b/build.js
@@ -110,14 +110,14 @@ const compileTestCode = (test, prefix = '', ownerPrefix = '') => {
       test.owner.slice(1) : test.owner;
 
   if (test.property == 'constructor') {
-    return `"bcd.testConstructor("${ownerAsProperty}")`;
+    return `bcd.testConstructor("${ownerAsProperty}")`;
   }
   if (test.owner === 'CSS.supports') {
     const thisPrefix = prefix ? `-${prefix}-` : '';
     return `CSS.supports("${thisPrefix}${test.property}", "inherit")`;
   }
   if (test.property.startsWith('Symbol.')) {
-    return `""Symbol" in self && "${test.property.replace('Symbol.', '')}" in Symbol && ${test.property} in ${ownerAsProperty}.prototype`;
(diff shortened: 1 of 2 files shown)


vinyldarkscratch:
Hey @foolip, I just had a thought about this.  Wouldn't it be better for the data to return `false` in this case, rather than `null`, and produce changes we can catch and verify?

One thing that just came to mind is that if a subfeature was implemented at the same time as the interface (say, Chrome 16), the collector wouldn't update BCD with a new version if the interface was implemented earlier (since Chrome 15 would result in `null`, telling update-bcd to return `true` for the version added, since it's the same as not knowing).

foolip:
@vinyldarkscratch can we instead tech update-bcd.js script to interpret the results correctly in situations like these? It's not all (current) null&rarr;true transitions due to missing interface objects that mean the subfeature was added at that point, but all false&rarr;true transitions should, modulo bugs. By turning those null&rarr;true into false&rarr;true, I'd trust all false&rarr;true transitions less.

The naive approach would be to let the update script infer false for all subfeatures in BCD if we've determined the parent feature to be supported, regardless of whether we've even tested for those subfeatures. (That would effectively be automating the changes to appease the lint.)

Are there other considerations I'm not seeing or giving the appropriate weight?

vinyldarkscratch:
That makes sense, to have the updater script account for the parent version to interpret results.  We'll probably also need to catch for the new importing feature of custom tests as well?

foolip:
I'm not sure what precise problem you foresee with custom tests, but if we can keep their true/null/false behavior close to what the generated tests do, then hopefully the updater code doesn't need to differentiate default tests from custom tests.

vinyldarkscratch:
With the new importing syntax, we're basically defining dependencies between APIs -- for example, an `AudioBufferSourceNode` requires an `AudioContext` to be created.  Thus, if we're preferring our code to throw an exception (causing a `null` result), we need to account for the version when `AudioContext` is supported the same way we would with subfeatures and their parents.

vinyldarkscratch:
Copied the conversation over to #506 so we can discuss this more in its own issue!

PR #497, opened 2020-09-16
Split the rest of non-standard.idl into separate files

foolip:
Plenty of small tinkering was done. api.CSSTransition.CSSTransition is
the only removed test, because the constructor shouldn't be there.

There are the added tests:
 - api.BlobBuilder.*
 - api.DataTransfer.getFiles
 - api.DataTransfer.getFilesAndDirectories
 - api.Element.scrollTopMax
 - api.UserProximityEvent.*
 - api.Window.ondevicelight
 - api.Window.ondeviceproximity
 - api.Window.onuserproximity
 - api.Window.routeEvent


foolip:
@vinyldarkscratch you might want to have a look at this after the fact, but I'll merge now since I have other changes I'd like to build on top of this.

vinyldarkscratch:
Thanks @foolip!  The files all look good to me.  😉 

Question: why are we removing the constructor for CSSTransition?  Is it because it shouldn't have one?  (If so, we should probably remove the entry from BCD.)

foolip:
@vinyldarkscratch that's right, `CSSTransition` shouldn't have a constructor. I missed that it was in BCD, sorry! http://mdn-bcd-collector.appspot.com/tests/api/CSSTransition/CSSTransition is still there, though, if you'd like to test for its non-support and remove it. We can also add it back, of course, with a comment.

vinyldarkscratch:
All good, thanks for confirming!  I checked and it seems Chrome was inaccurately marked as having a constructor, so we can safely remove it from BCD.  I'll submit a PR and request your review!

PR #502, opened 2020-09-17
Custom tests import support

### LINKED ISSUE #500: Custom tests: import other custom tests?
One thing I noticed is that a lot of APIs I've had to write custom tests for depend on other APIs (such as how the `AudioNode` depends on `AudioContext`).  As a result, there's a lot of repeated code.  We should implement a form of importing from other tests to reduce the amount of duplicate code.

I'm thinking of a syntax roughly along the lines of `<%ident:varname%>`, where the `ident` is the identifier to copy from, and the `varname` is what to rename the `instance` variable to.

vinyldarkscratch:
This PR fixes #500 by implementing a syntax to import the custom test code from other APIs and their members with a simple syntax of `<%ident:varname%>`, where the `ident` is the identifier to copy from, and the `varname` is what to rename the `instance` variable to.  This also updates the audio node code to use this new syntax.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -118,6 +118,36 @@ bcd.addTest('css.properties.custom-property', "(function() {return CSS.supports(
 
 Tips: make sure that all return statements will return a boolean, and implement thorough feature checking.
 
+#### Importing code from other tests
+
+Sometimes, some features will depend on the setup and configuration from other features, especially with APIs.  To prevent repeating the same code over and over again, you can import code from other custom tests to build new ones quicker.  The syntax to specify a test import is the following: `<%ident:varname%>`, where `ident` is the full identifier to import from, and `varname` is what to rename the `instance` variable from that test to.
+
+Example:
+
+The following JSON...
+
+```
+{
+  "api": {
+    "AudioContext": {
+      "__base": "var instance = new (window.AudioContext || window.webkitAudioContext)();"
+    },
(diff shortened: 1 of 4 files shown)


DESIGN.md
```diff
@@ -118,6 +118,36 @@ bcd.addTest('css.properties.custom-property', "(function() {return CSS.supports(
...
+The following JSON...
+
+```
+{
+  "api": {
+    "AudioContext": {
+      "__base": "var instance = new (window.AudioContext || window.webkitAudioContext)();"
+    },
+    "AudioDestinationNode": {
+      "__base": "<%api.AudioContext:audioCtx%> if (!audioCtx) {return false}; var instance = audioCtx.destination;"
```
foolip, line 136: This is just an example, but the `if (!audioCtx) {return false}` makes me nervous about trusting the results just like the issue with assuming interface objects exist. There might be many ways to get an instance of this, and not finding one isn't very conclusive evidence that the subfeature isn't supported. Also, wouldn't the code throw an exception more often than not?

Just assuming that `audioCtx` was set and letting the code fail if not would be preferable, I think.

vinyldarkscratch: We can certainly do that, though we should implement, in update-bcd, a piece of code to check to see if `AudioContext` is supported and compare it to the `null` values.  I've mentioned this in #506!

PR #504, opened 2020-09-17
Validate that all types in Web IDL are known

foolip:
Fix some of the simpler cases, and leave TODOs for the rest


vinyldarkscratch (approved):
Looking good to me, thanks!

Issue #488, opened 2020-09-14
Formatting for non-JavaScript files


queengooborg:
Inspired by #485.  ESLint is great for formatting the JavaScript files, but for all the others, it has no support.  We should find and introduce a package that can format our other files, like the Markdown, JSON, YAML, and EJS files.

In my research, I found a package called [Unibeautify](https://unibeautify.com/) that has support for many different beautifiers and many different languages.  It may be something to look into to simplify our code/file formatting!  It also links to the beautifiers it has support for, and has a matrix of the languages it supports to the beautifiers that can handle them: https://unibeautify.com/docs/languages

vinyldarkscratch:
#509 and #510 have taken care of the Markdown, JSON, and YAML, leaving only the EJS files.  Unibeautify __does__ seem like a nice project (I'll probably end up contributing to it in my spare time, haha), but it's got some big flaws to work out before it can be used.  Perhaps the individual linters it works with may be worth looking into...

PR #505, opened 2020-09-17
Add linebreaks to custom tests for better display

vinyldarkscratch:
This PR adds linebreaks to the custom tests to improve their display, making them more legible, like so:

Screen Shot 2020-09-17 at 11 55 29

### DIFF
--- a/build.js
+++ b/build.js
@@ -40,7 +40,8 @@ const getCustomTestAPI = (name, member) => {
   let test = false;
 
   if (name in customTests.api) {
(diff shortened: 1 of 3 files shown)


foolip:
Good stuff! I've also wondered if maybe we could run prettier on the snippets, so one could just create a long snippet and not worry about formatting at all.

vinyldarkscratch:
I was thinking of the same thing, though I didn't want to add a dependency just for that.  I was thinking about presenting additional code formatters in #488, however, so might as well do it altogether!

PR #535, opened 2020-09-25
Improve update-bcd

### LINKED ISSUE #506: update-bcd: account for feature's parent/dependency support
@vinyldarkscratch can we instead teach update-bcd.js script to interpret the results correctly in situations like these? It's not all (current) null&rarr;true transitions due to missing interface objects that mean the subfeature was added at that point, but all false&rarr;true transitions should, modulo bugs. By turning those null&rarr;true into false&rarr;true, I'd trust all false&rarr;true transitions less.

The naive approach would be to let the update script infer false for all subfeatures in BCD if we've determined the parent feature to be supported, regardless of whether we've even tested for those subfeatures. (That would effectively be automating the changes to appease the lint.)

Are there other considerations I'm not seeing or giving the appropriate weight?

_Originally posted by @foolip in https://github.com/foolip/mdn-bcd-collector/pull/492#issuecomment-694180857_

vinyldarkscratch:
This PR provides a number of updates to the `update-bcd` script to improve its performance, including the following:

- If a version number in BCD doesn't contradict a generated range, don't replace it with the range
- Remove a version_removed if proven wrong
- Account for the support of the parent (fixes #506)

### DIFF
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -310,6 +310,8 @@ const update = (bcd, supportMatrix) => {
           simpleStatement.version_added = inferredStatement.version_added;
           modified = true;
         }
+      } else if (simpleStatement.version_removed) {
+        delete simpleStatement.version_removed;
       }
     }
   }
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -284,9 +284,9 @@ const update = (bcd, supportMatrix) => {
         inferredStatement.version_added.includes('≤')
       ) {
         if (compareVersions.compare(
-          simpleStatement.version_added.replace('≤', ''),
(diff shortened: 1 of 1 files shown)


foolip:
Thanks for the updates, @vinyldarkscratch!

This is tricky business, so of all the code in this repo, I think this logic of combining existing support statements with inferred support statements might be most worth unit testing. If you do that, I'd very much like to review.

vinyldarkscratch:
Certainly!  I've been thinking about adding unittests for update-bcd soon, though considering the more complicated logic, it's probably a good idea to get on it as soon as possible.

PR #538, opened 2020-09-25
Don't include utils.js in production

vinyldarkscratch:
This PR reverts a change to use utils.js in the GitHub module, fixing production.


vinyldarkscratch:
Good news, this fixed the issue in production!

PR #542, opened 2020-09-25
Use fs-extra's writeJson over own utils.writeFile

foolip:
We already depend on fs-extra, and in the places where we used
utils.writeFile do have any special requirements about the formatting,
so switch those to fs-extra's writeJson.

Fold what remains of utils.js into update-bcd.js.


vinyldarkscratch (approved):
It looks like utils.js is being removed in this -- it's a shame as I was hoping to place commonly-used functions, but I guess there's not really any that we need to worry about right now.  Looks good!

foolip:
Right, we weren't left with anything used in more than one place. As soon as we need it, we can add it back.

PR #547, opened 2020-09-26
Force LF when checking out

saschanaz:
Git for Windows uses CRLF by default while the eslint configuration requires LF. This makes sure git uses LF regardless of local git configuration.


foolip (approved):
I haven't tested this, but trust that it works!

PR #548, opened 2020-09-26
Port bash scripts to JS

### LINKED ISSUE #546: `npm prepare` does not run on Windows
... as it uses bash statement.

saschanaz:
Bit longer but better portability.

Closes #546

### DIFF
--- a/scripts.js
+++ b/scripts.js
@@ -1,0 +1,22 @@
+const childProcess = require('child_process');
+
+const exec = (cmd, env) => {
+  console.log(`> ${cmd}`);
+  return childProcess.execSync(cmd, {env, stdio: 'inherit'});
+};
+
+const prepare = () => {
+  try {
+    process.chdir('node_modules/puppeteer');
+  } catch {
+    return;
+  }
+  exec('node install.js', {PUPPETEER_PRODUCT: 'firefox'});
+};
+
+const command = process.argv[2];
+if (command === 'prepare') {
+  prepare();
+} else if (command === 'unittest') {
+  exec('nyc mocha --recursive unittest', {NODE_ENV: 'test'});
+}
--- a/scripts.js
+++ b/scripts.js
@@ -1,3 +1,17 @@
+// Copyright 2020 Mozilla Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 const childProcess = require('child_process');
 
 const exec = (cmd, env) => {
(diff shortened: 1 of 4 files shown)


foolip:
I've opened https://github.com/foolip/mdn-bcd-collector/pull/549 to run CI on macOS and Windows as well.

saschanaz:
You want me to rebase this on it?

scripts.js
```diff
@@ -0,0 +1,23 @@
+const childProcess = require('child_process');
```
foolip, line 1: Can you add a copyright header to this file, like the others? I guess it would be either "Copyright 2020 Mozilla Corporation" or "Copyright 2020 Mozilla Foundation", which I don't know.

foolip (commented):
Thanks for fixing this! Just one paperwork nit :)

foolip:
@saschanaz I've landed https://github.com/foolip/mdn-bcd-collector/pull/549 with the Windows config commented out, so that you can enable it in this PR to show that it's working.

scripts.js
saschanaz: I think the MoCo one should work, but I should probably ask my manager about this. Making this a draft until I figure it out...

foolip: Hope it's not a whole process, if so I apologize! :)

saschanaz:
My local unittest result on Windows:

```
> npm run unittest

.... skipped, just normal successful results ....

  1) "after all" hook in "{root}"

  129 passing (7s)
  1 failing

  1) "after all" hook in "{root}":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)



[Error: EBUSY: resource busy or locked, unlink 'C:\Users\Kagami\AppData\Local\Temp\puppeteer_dev_firefox_profile-AC0rlQ\cert9.db'] {
  errno: -4082,
  code: 'EBUSY',
  syscall: 'unlink',
  path: 'C:\\Users\\Kagami\\AppData\\Local\\Temp\\puppeteer_dev_firefox_profile-AC0rlQ\\cert9.db'
}
```

Then it halts forever. 🤔🤔

saschanaz:
I guess it's a puppeteer issue: https://github.com/puppeteer/puppeteer/issues/5673

unittest/puppeteer/harness.js
```diff
@@ -22,7 +22,12 @@ const path = require('path');
 
 const {app} = require('../../app');
 
-const products = ['chrome', 'firefox'];
+const products = ['chrome'];
+if (process.platform !== 'win32' || Date.now() > Date.parse('2020-10-30')) {
```
foolip, line 26: This will probably begin failing without the code changing. If just disabling on Windows with a TODO isn't palatable, then can you instead make this depend on the version of puppeteer, from `require('../../package.json').devDependencies.puppeteer`? I'd rather have it fail on upgrading Puppeteer than based on the calendar.

Also, to make it more visible that it's disabled, you might try `this.skip()` as described in https://mochajs.org/#inclusive-tests

scripts.js
saschanaz: It turns out that Copyright 2020 MoCo is indeed correct here. 👍

vinyldarkscratch (approved):
Looking all good, thanks @saschanaz!

scripts.js
foolip: Thanks for sorting that out!

PR #550, opened 2020-09-26
Add SharedWorker support

saschanaz:
I went the lazy way and just copied runWorker, but I should probably merge those two. Or do you want this as-is, as it seems runServiceWorker is also quite duplicated?

### DIFF
--- a/static/resources/serviceworker.js
+++ b/static/resources/serviceworker.js
@@ -1,4 +1,4 @@
-// Copyright 2020 Mozilla Corporation
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/static/resources/serviceworker.js
+++ b/static/resources/serviceworker.js
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2020 Mozilla Corporation
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/static/resources/sharedworker.js
+++ b/static/resources/sharedworker.js
@@ -1,0 +1,34 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


views/index.ejs
```diff
@@ -17,6 +17,7 @@
       <option value="">All Global Exposures</option>
       <option value="?exposure=Window">Window</option>
       <option value="?exposure=Worker">Worker</option>
```
vinyldarkscratch, line 19: Since we're here, we might as well specify this is a dedicated worker?

```suggestion
      <option value="?exposure=Worker">Dedicated Worker</option>
```

views/index.ejs
```diff
@@ -17,6 +17,7 @@
...
+      <option value="?exposure=SharedWorker">SharedWorker</option>
```
vinyldarkscratch, line 20:
```suggestion
      <option value="?exposure=SharedWorker">Shared Worker</option>
```

static/resources/serviceworker.js
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2020 Mozilla Corporation
```
vinyldarkscratch, line 1: Can we undo this change?  This is the only change in this file -- I'm guessing that this was the wrong file.

vinyldarkscratch (changes requested):
Thanks @saschanaz, this looks good to me!  Just got a couple of changes to request.

static/resources/sharedworker.js
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 Mozilla Corporation
```
saschanaz, line 1: I'll wait until I figure it out about the license: https://github.com/foolip/mdn-bcd-collector/pull/548#discussion_r495488389

Or should this still be Google LLC as this is really just a copy of worker.js? 👀

vinyldarkscratch: Personally I would think that the license would remain the same since the new file is a carbon copy of the existing file?  (I'm requesting a revert of this change because it's modifying the service worker file, not the new shared worked.  😛)

saschanaz: Okay, I'll do that 👍

saschanaz:
BTW, is it intentional that "Limit Exposure" works only when selecting a specific test but not when doing "All Tests"?

vinyldarkscratch (approved):
Awesome, thank you for the quick changes, we're now all set to go!  👍 

In response to your question about the scope limitation not working on "All Tests", it looks like you've found a bug -- the scope _should_ be limiting, but it doesn't seem like it's getting added into the URL for some odd reason.  I'm working on the views, so I'll tackle it ASAP!

PR #553, opened 2020-09-27
Update unittests

vinyldarkscratch:
This PR improves the unittests within the repository, including adding unittests for `find-missing` and `update-bcd`.


unittest/unit/find-missing.js
```diff
@@ -0,0 +1,108 @@
...
+
+'use strict';
+
+const assert = require('chai').assert;
+
+const proxyquire = require('proxyquire');
+
+const tests = {
+  'api.AbortController': {
+    tests: [{code: '"AbortController" in self', prefix: ''}],
```
foolip, line 23: find-missing.js uses `Object.key(tests)` so there's no need to maintain the detailed structure in mock data. Instead, just `'api.AbortController': {}` or `'api.AbortController': 'dummy'` would do. Something more like the dummy bcd data.

unittest/unit/find-missing.js
```diff
@@ -0,0 +1,108 @@
...
+      },
+      'font-face': {
+        __compat: {}
+      }
+    }
+  }
+};
+
+const {traverseFeatures, findMissing} = proxyquire('../../find-missing', {
+  './tests.json': Object.assign({}, tests),
```
foolip, line 83: Why the `Object.assign` here? `tests` isn't used elsewhere, so it's not to make a copy and avoid interaction between tests. If it were, then a deep copy would be more appropriate.

unittest/unit/update-bcd.js
```diff
@@ -14,13 +14,330 @@
 
 'use strict';
 
-const assert = require('assert');
-const updatebcd = require('../../update-bcd');
+const chai = require('chai');
```
foolip, line 17:
```suggestion
const {assert,expect} = require('chai');
```

unittest/unit/update-bcd.js
```diff
@@ -14,13 +14,330 @@
...
+    it('Samsung Internet (12.1)', () => {
+      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (Linux; Android 11; Pixel 2) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/12.1 Chrome/79.0.3945.136 Mobile Safari/537.36', bcd.browsers), ['samsunginternet_android', '12.1']);
+    });
+
+    it('Yandex Browser (not in BCD)', () => {
+      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (Windows NT 6.3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 YaBrowser/17.6.1.749 Yowser/2.5 Safari/537.36', bcd.browsers), [null, null]);
+    });
+  });
+
+  describe('getSupportMap', () => {
```
foolip, line 316: Can you also test `getSupportMatrix`? That's fairly important in terms of managing conflicting results for the same version, and seems to not actually do it in a deliberate way, instead the last report wins, it seems.

vinyldarkscratch: Of course, adding those tests right now actually!  (This PR's still in drafting stages.)

unittest/unit/find-missing.js
vinyldarkscratch: What's odd is that without the `Object.assign`, it seems to merge both the dummy object and the real JSON file, though I'm not sure why.  Wondering if it's a `proxyquire` bug...

unittest/unit/update-bcd.js
```diff
@@ -14,13 +14,728 @@
...
+      }).to.throw('result not true/false/null; got 87');
+    });
+  });
+
+  describe('update', () => {
+    const supportMatrix = getSupportMatrix(bcd.browsers, reports);
+    let bcdCopy;
+
+    beforeEach(() => {
+      bcdCopy = Object.assign({}, bcd);
```
foolip, line 639:
```suggestion
      bcdCopy = JSON.parse(JSON.stringify(bcd));
```

`Object.assign` results in a shallow copy, would would end up with interactions between tests. Using `JSON.parse(JSON.stringify(bcd))` is a quick if not pretty way to make a copy of a simple object.

unittest/unit/update-bcd.js
```diff
@@ -14,13 +14,728 @@
...
+
+    it('Yandex Browser (not in BCD)', () => {
+      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (Windows NT 6.3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 YaBrowser/17.6.1.749 Yowser/2.5 Safari/537.36', bcd.browsers), [null, null]);
+    });
+  });
+
+  describe('getSupportMap', () => {
+    it('normal', () => {
+      assert.deepEqual(getSupportMap(reports[0]), new Map([
+        ['api.AbortController', {result: true, prefix: ''}],
```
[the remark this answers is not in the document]
vinyldarkscratch: There was some reason behind it, but I can't remember exactly what it was.  I think it was improved map parsing?

unittest/unit/update-bcd.js
```diff
@@ -395,12 +381,20 @@ describe('BCD updater', () => {
...
+      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.125 Safari/537.36 Edg/84.0.522.59', bcd.browsers), ['edge', '84']);
+    });
+
     it('Safari', () => {
       assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0 Safari/605.1.15', bcd.browsers), ['safari', '14']);
     });
 
     it('Safari iOS', () => {
-      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (iPhone; CPU iPhone OS 14_0_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0 Mobile/15E148 Safari/604.1', bcd.browsers), ['safari_ios', '14']);
+      assert.deepEqual(getBrowserAndVersion('Mozilla/5.0 (iPhone; CPU iPhone OS 13_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.1 Mobile/15E148 Safari/604.1', bcd.browsers), ['safari_ios', '']);
```
foolip, line 393: Aha, so `getBrowserAndVersion` actually doesn't handle Safari iOS correctly. I've filed https://github.com/foolip/mdn-bcd-collector/issues/562.

foolip (approved):

unittest/unit/update-bcd.js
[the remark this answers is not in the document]
foolip: Alright, let's leave that alone to get these tests landed.

unittest/unit/find-missing.js
vinyldarkscratch: Found the issue: it looks like `proxyquire` has call-thru capabilities, so that anything that isn't mocked in `proxyquire` is obtained from the original dependency.  That's why BCD is being loaded, and in fact, JSON files as well in all other tests.

I opened #594 to resolve this.

PR #560, opened 2020-09-29
Use logger.* over console.* in selenium.js and update-bcd.js

foolip:
This is to suppress log output while running tests:
https://github.com/foolip/mdn-bcd-collector/pull/553


vinyldarkscratch (approved):
Looks good to me, thanks!

PR #561, opened 2020-09-29
Use fast-deep-equal over own isEquivalent helper

foolip:
We already depend on it via eslint.


vinyldarkscratch (approved):
This PR looks good to me, thanks!  👍

Issue #562, opened 2020-09-29
Safari iOS version not detected correctly


foolip:
https://github.com/foolip/mdn-bcd-collector/blob/36b42144b357976f0919083b6b2425629d50acb0/update-bcd.js#L25-L53

We have two bugs here:
- Safari for iOS should use the OS version, per https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#safari-for-ios-versioning
- If BCD doesn't have a release for 10.2, but does have 10.1 and 10.3, we should detect the version as 10.1. BCD skips some versions that exist, but which had no noteworthy changes. This also affects Samsung Internet, per the TODO.

PR #565, opened 2020-09-29
Revamp getBrowserAndVersion to handle Safari iOS and Samsung Internet

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/562.


vinyldarkscratch (approved):
Sweet, thank you!

PR #573, opened 2020-10-01
Fix the IIRFilterNode instance factory (required arguments)

foolip:


foolip:
@vinyldarkscratch this was the only test returning null in Chrome and Firefox results. In Safari, `ConstantSourceNode` and `StereoPannerNode` also throw, but I haven't looked into those yet. It seems like a decent goal to have no inconclusive tests in the most recent version of each browser.

PR #594, opened 2020-10-06
Prevent callThru on proxyquire

vinyldarkscratch:
This PR prevents `proxyquire` from loading the original files or modules, thus resolving the issue mentioned in https://github.com/foolip/mdn-bcd-collector/pull/553#discussion_r495793082.


foolip (commented):
Cool, thanks for finding a solution!

PR #596, opened 2020-10-06
Suggest similar ident if test not found

### LINKED ISSUE #582: Case-insensitive URLs?
A very little thing, but it might be nice to have case-insensitive URLs for the tests?  Or more accurately, redirects the URL to the proper casing.  (I've been manually typing out the URLs quite a bit lately and end up accidentally forgetting to capitalize the API name.)

vinyldarkscratch:
This PR introduces an updated test 404 page which includes a "Did you mean" section to suggest the correct ident when there's a typo, including caps issues.  This fixes #582.

### DIFF
--- a/views/testnotfound.ejs
+++ b/views/testnotfound.ejs
@@ -1,0 +1,9 @@
+<%- contentFor('body') %>
+<h1>Test Not Found</h1>
+<p>Could not find a test for <code><%= ident %></code>. 
+  <% if (suggestion) { %>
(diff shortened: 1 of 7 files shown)


views/testnotfound.ejs
```diff
@@ -0,0 +1,9 @@
+<%- contentFor('body') %>
+<h1>Test Not Found</h1>
+<p>Could not find a test for <code><%= ident %></code>. 
+  <% if (suggestion) { %>
+    <span class="notice">Did you mean <a href="/tests/<%= suggestion.replace(/\./, '/') %>"><code><%= suggestion %></code></a>?</span>
```
foolip, line 5: The replace call here will only replace the first period, I think.

Issue #406, opened 2020-09-03
Interface member tests should be inconclusive if the interface object is not exposed


foolip:
http://mdn-bcd-collector.appspot.com/api/Element/getAttribute/ has a generated test `"Element" in self && "getAttribute" in Element.prototype`. That should be just `"getAttribute" in Element.prototype` so that it throws an exception if Element was not exposed. The result should be null,  not false.

The reason for this is that if the interface object isn't exposed, we can't be sure that the member isn't. The interface might still be supported, just not exposed, or exposed under an alias which is already in the data.

vinyldarkscratch:
Actually, I had made that change intentionally to produce `false` if the interface isn't exposed normally.  This lets us write more data in BCD as `false` if the interface is not implemented, allowing us to fix consistency issues.  If the interface isn't exposed normally and instead is accessible through other ways, that's where we should be writing a custom test.

foolip:
My concern is that this will cause us to suggest updates in BCD that are wrong because the interface object is missing. https://chromium.googlesource.com/chromium/src/+/b7e3a7693e9f76ecedf3c78e9484463259d7f4b8 is a typical example of a feature history that would be problematic. The `RTCDTMFSender` interface object wasn't exposed for a long time, but it was implemented. By testing for the interface, we would suggest a lot of changes to subfeatures of https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender. From the fact that the interface object has gone from not exposed to exposed, I think at most we should revisit if the main feature has the right version in BCD. BCD lints would catch if one forgets to also update the subfeatures.

One could of course handle this in update-bcd.js and make the updating of subfeatures depend on the outcome of the status of the main feature. However, if the end result is the same, I'd favor the simplicity of shorter tests and making it explicit that the outcome of the test was inconclusive because the interface object is missing.

foolip:
Is this something we can resolve sooner rather than later? When reviewing https://github.com/mdn/browser-compat-data/pull/6642, my confidence in any change that changes an interface member to false is not high, because it might be a side effect of this issue. Even when it's been verified manually, if the manual testing made the same assumptions, then the main point of review is to consider if the interface itself has been exposed under a different name, or had `[NoInterfaceObject]` at some point.

By treating these differently, review will be faster overall, because cases where the interface object and some members exist, but others do not, are ones where one can be highly confident in the change and need not spend much time probing the history of the feature.

foolip:
[...]

PR #607, opened 2020-10-09
Make custom tests inconclusive (null) if any dependency is missing

foolip:
By testing for dependencies and returning false, one could not tell from
the results if the important part of the test had run, or if it was some
precondition that had failed.

Where the code will necessarily throw an exception if the precondition
isn't met, just let the error be thrown. In the few cases where the
value isn't treated as an object, instead explicitly return null.

Related: https://github.com/foolip/mdn-bcd-collector/issues/406


vinyldarkscratch:
I don't think we're ready for this quite yet, until `update-bcd` has been refactored to account for imports in custom tests?

foolip:
You're right, we'd need some notion of dependencies between tests which are trivial for subfeatures but not so in this case.

PR #609, opened 2020-10-11
Selenium upgrades: multiple OS support

vinyldarkscratch:
This PR upgrades the Selenium script to enhance testing between OSes and OS versions, by testing browsers in both Windows and macOS (aside from Safari which will only run in macOS), as well as testing in the most recent Windows versions and falling back to older versions when needed.

### DIFF
--- a/selenium.js
+++ b/selenium.js
@@ -102,7 +102,7 @@ const run = async (browser, version, os) => {
       Capability.BROWSER_NAME,
       Browser[browser.toUpperCase()]
   );
-  capabilities.set(Capability.VERSION, version);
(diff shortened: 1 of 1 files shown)


selenium.js
```diff
@@ -234,11 +282,17 @@ if (require.main === module) {
               describe: 'Limit the browser(s) to test',
               type: 'string',
               choices: ['chrome', 'edge', 'firefox', 'ie', 'safari']
+            })
+            .option('os', {
+              describe: 'Limit the os(es) to test',
```
foolip, line 287: How to control which versions of each are tested?

PR #619, opened 2020-10-14
Add MerchantValidationEvent custom IDL

foolip:


foolip:
We'll need this after `MerchantValidationEvent` goes away with a webref upgrade.

PR #623, opened 2020-10-17
Expose reusable instances globally

vinyldarkscratch:
This PR exposes the `reusableInstances` variable to make it easier to debug issues in the console.


foolip:
Putting it under `bcd`, the (hopefully) only other thing added to the global object, would be nice.

PR #633, opened 2020-10-20
Post results before displaying them

### LINKED ISSUE #632: Unable to upload results on Chrome for Android
This is https://github.com/foolip/mdn-bcd-collector/issues/486 again, but failing on Chrome for Android:

The console error is:
```
Uncaught TypeError: Cannot read property 'replace' of undefined
    at report (harness.js:436)
    at harness.js:351
    at MessagePort.messageChannel.port1.onmessage (harness.js:293)
```

So the failure is here:
https://github.com/foolip/mdn-bcd-collector/blob/743d92c1ee0a61061d52e299b3cd0a8cac0af7fc/static/resources/harness.js#L436

Presumably the fix is simple. We should also have some kind of regression testing in place so that it doesn't keep breaking on mobile.

vinyldarkscratch:
This PR fixes #632.

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -464,11 +464,13 @@
         resultNameEl.appendChild(resultValueEl);
         thisResultEl.appendChild(resultNameEl);
 
-        var codeEl = document.createElement('code');
-        codeEl.innerHTML = result.info.code.replace(/\n/g, '<br />');
-        thisResultEl.appendChild(codeEl);
-        thisResultEl.appendChild(document.createElement('br'));
-        thisResultEl.appendChild(document.createElement('br'));
+        if (result.info.code) {
+          var codeEl = document.createElement('code');
+          codeEl.innerHTML = result.info.code.replace(/\n/g, '<br />');
+          thisResultEl.appendChild(codeEl);
+          thisResultEl.appendChild(document.createElement('br'));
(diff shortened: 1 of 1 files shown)


static/resources/harness.js
```diff
@@ -402,6 +402,38 @@
...
+        return;
+      }
+
+      client.open('POST', '/api/results?for='+encodeURIComponent(location.href));
+      client.setRequestHeader('Content-Type', 'application/json;charset=UTF-8');
+      client.send(body);
+      client.onreadystatechange = function() {
+        if (client.readyState == 4) {
+          if (client.status >= 200 && client.status <= 299) {
+            updateStatus('Results uploaded. <a href="/results" id="submit">Submit to GitHub</a>');
```
foolip, line 426: Shouldn't this link to the created PR?

foolip (commented):
Thanks for fixing this!

static/resources/harness.js
vinyldarkscratch: In this step, the results are only submitted to the server and haven't been uploaded to GitHub.  This is to allow the Selenium script to run the test without creating a monstrous amount of PRs and download the results afterwards.

foolip: Oh right, I forgot how this all works :D

PR #639, opened 2020-10-21
Add blacklist functionality to disable certain tests

vinyldarkscratch:
This PR adds a `blacklist` parameter to the test running URL to disable specific tests when running all of them.  This is to allow us to disable certain APIs for Chrome versions when they crash the browser, so we can still obtain other data for the other APIs.

### DIFF
--- a/app.js
+++ b/app.js
@@ -176,7 +176,8 @@ app.all('/tests/*', (req, res) => {
       title: `${ident || 'All Tests'}`,
       layout: false,
       tests: foundTests,
-      hideResults: req.query.hideResults
+      hideResults: req.query.hideResults,
+      blacklist: (req.query.blacklist ? req.query.blacklist.split(',') : [])
     });
   } else {
     res.status(404).render('testnotfound', {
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -176,7 +176,8 @@ app.all('/tests/*', (req, res) => {
       title: `${ident || 'All Tests'}`,
       layout: false,
       tests: foundTests,
-      hideResults: req.query.hideResults
+      hideResults: req.query.hideResults,
+      blacklist: (req.query.blacklist ? req.query.blacklist.split(',') : [])
```
foolip, line 180: Can we call this `exclude` or `ignore`? It'll be in URLs and in the web standards space most instances of "blacklist" have been replaced by "blocklist", "ignorelist", or similar over the past year or so, to not link white/black to good/bad. In almost every case I think it's also been clearer after the change.

vinyldarkscratch: Ahhhh, I wasn't aware that the terms blacklist/whitelist are phasing out, but I can definitely see why!  Following up now!

PR #641, opened 2020-10-22
Fix tests for HTML element APIs

vinyldarkscratch:
This PR fixes the HTML element API tests by testing if the constructor is not `HTMLUnknownElement` __or__ `HTMLElement` (before, it only tested if it's not `HTMLUnknownElement`, which causes issues when the browser doesn't support `HTMLUnknownElement`).

### DIFF
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -205,232 +205,232 @@
       "__base": "if (!reusableInstances.audioContext) {return false;} var instance = reusableInstances.audioContext.createGain();"
     },
     "HTMLAnchorElement": {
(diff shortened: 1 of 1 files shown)


custom-tests.json
```diff
@@ -205,232 +205,232 @@
       "__base": "if (!reusableInstances.audioContext) {return false;} var instance = reusableInstances.audioContext.createGain();"
     },
     "HTMLAnchorElement": {
```
foolip, line 207: Can't we instead do `if (instance.constructor.name !== 'HTMLAnchorElement') { return false; }`? Or are the very old browsers where `instance.constructor.name` is the wrong thing?

vinyldarkscratch: Taken care of in #644!

PR #644, opened 2020-10-22
HTML/SVG elements: improve API confirmation

vinyldarkscratch:
This PR changes the HTML and SVG element API constructor confirmation to make sure the API names are exclusively the expected ones, rather than if it's not `HTMLUnknownElement`, `HTMLElement`, or `SVGElement`.

### DIFF
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -205,49 +205,49 @@
       "__base": "if (!reusableInstances.audioContext) {return false;} var instance = reusableInstances.audioContext.createGain();"
     },
     "HTMLAnchorElement": {
-      "__base": "var instance = document.createElement('a'); if (instance.constructor.name =!= 'HTMLAnchorElement') {return false;}"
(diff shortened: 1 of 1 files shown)


custom-tests.json
```diff
@@ -205,232 +205,232 @@
       "__base": "if (!reusableInstances.audioContext) {return false;} var instance = reusableInstances.audioContext.createGain();"
     },
     "HTMLAnchorElement": {
-      "__base": "var instance = document.createElement('a'); if (instance.constructor.name === 'HTMLUnknownElement' || instance.constructor.name === 'HTMLElement') {return false;}"
+      "__base": "var instance = document.createElement('a'); if (instance.constructor.name =!= 'HTMLAnchorElement') {return false;}"
```
foolip, line 208: This diff is delightful, I gotta suspect an `!` was inserted at the same place in many lines, and it didn't quite work out :D

vinyldarkscratch: It was, something odd happened with the multiple selection on my editor.  Fixed!

PR #655, opened 2020-10-23
Show app version on homepage

vinyldarkscratch:
This PR includes the app version in the title section on the homepage.  Useful to ensure a production version is successfully deployed.

Screen Shot 2020-10-23 at 16 03 40


foolip:
Nice, I've wanted this too!

PR #686, opened 2020-10-28
Overhaul page styling

vinyldarkscratch:
This PR performs a big update to the page styling, including added icons, slight gradients for texture, a brand new favicon.

Note: this _does_ hook up the layout onto the tests page, which includes the stylesheet we're using everywhere else (thus, it doesn't load the stylesheet after the tests anymore).  The reasoning behind this is: on some older browsers, the switch of the font causes quite a bit of lag when there's lots of results to display, and since users will be visiting the homepage, the stylesheet will be cached so load time is not impacted by internet speed.

Screen Shot 2020-10-28 at 02 58 40
Screen Shot 2020-10-28 at 03 00 55
Screen Shot 2020-10-28 at 03 01 13
Screen Shot 2020-10-28 at 03 01 27


foolip:
@vinyldarkscratch I love it! Do you plan to make further changes, or should I give this a quick review now?

vinyldarkscratch:
💜

I have no changes planned at the moment, feel free to review and merge whenever you'd like!

foolip:
@vinyldarkscratch can we limit the width of the "known caveats" box and have it centered like the stuff above it?

vinyldarkscratch:
Giving it some more thought, I just had two ideas for things I'd like to get in before deploying this, so I'm going to drop the version number bump.

1) I'm thinking it would be nice to add a little bit in the corner that displays the user's browser name and version, or "unsupported" if BCD doesn't record data for it.  This will be especially helpful for when I'm in VMs and forget which one I'm working with.
2) We should add more features to detect for this big release, so we can get collector coverage up.

Now that I'm thinking about it...maybe this is also the time for a 1.0 release?

foolip:
> I'm thinking it would be nice to add a little bit in the corner that displays the user's browser name and version, or "unsupported" if BCD doesn't record data for it.

That would be great, yeah. Sometimes I've been going https://www.whatsmyua.info/ in addition to the collector in some setting where I have to type out the URLs manually, and having the detected browser + version right there on the front page would be great.

> maybe this is also the time for a 1.0 release?

Let's do it! We won't run out of integers, so we can always go to 2.0 if we make another round of large changes 👍

vinyldarkscratch:
Got the UA parsing added to the styling, so that we'll always know what our script thinks the browser is, including warnings if it's not in BCD!

vinyldarkscratch:
Alrighty, I think it makes sense to merge this PR now, so that it'll be included in our upcoming 1.0 release!

foolip:
http://staging-dot-mdn-bcd-collector.appspot.com/ is live, and I like it! :)

vinyldarkscratch:
Great to know it works in staging!

PR #722, opened 2020-11-11
1.0.0

vinyldarkscratch:


vinyldarkscratch:
(Note: my request for review isn't because of some big change that this PR does; all it does is bump the version number.  The review request is more of a question of "are we ready to go on this version?")

vinyldarkscratch:
Actually...wait.  I want to try to get more custom tests, and more importantly, promise support, in.

PR #724, opened 2020-11-12
Add promise testing support (+ custom test for BatteryManager)

vinyldarkscratch:
This PR reworks harness.js and build.js to add support for testing promises and other asynchronous tests, reworking the test structure to use callbacks.  Additionally, this adds a custom test for BatteryManager to show proof of concept.

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -201,7 +201,7 @@
       try {
         var value = eval(test.code);
 
-        if ('then' in value) {
+        if ('Promise' in self && value instanceof Promise) {
           Promise.resolve(value).then(process);
         } else {
           process(value);
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -202,7 +202,7 @@
         var value = eval(test.code);
 
         if ('Promise' in self && value instanceof Promise) {
-          Promise.resolve(value).then(process);
+          Promise.resolve(value).then(process, process);
         } else {
           process(value);
         }
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -181,6 +181,16 @@
     var result = {name: data.name, info: {}};
 
     function runTest(i, callback) {
+      function process(value) {
+        processValue(value, data, test, result);
+        if (result.result === true) {
+          callback(result);
+          return;
+        } else {
+          runTest(i + 1, callback);
+        }
+      }
+
       if (i >= data.tests.length) {
         callback(result);
         return;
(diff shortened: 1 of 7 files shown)


static/resources/harness.js
```diff
@@ -144,59 +177,62 @@
   //
   // If the test doesn't return true or false, or if it throws, `result` will
```
foolip, line 178: Can you update this comment to say what `callback` is?

static/resources/harness.js
```diff
@@ -131,6 +131,39 @@
     return result;
   }
 
+  function processValue(value, data, test, result) {
+    if (value && typeof value === 'object' && 'result' in value) {
```
foolip, line 135: This is moved code, not new, but what's this for? Are some tests returning `{ result: true }` instead of just `true`?

static/resources/harness.js
```diff
@@ -144,59 +177,62 @@
...
   // If the test doesn't return true or false, or if it throws, `result` will
   // be null and a `message` property is set to an explanation.
-  function test(data) {
+  function test(data, callback) {
     var result = {name: data.name, info: {}};
 
-    for (var i = 0; i < data.tests.length; i++) {
+    function runTest(i, callback) {
+      function process(value) {
+        processValue(value, data, test, result);
```
foolip, line 185: Some comments to explain what's going on here would be great. How does a mix of sync and async tests work? Given only async tests, are they run one after the other or all in parallel? If they're run in series, what happens between one test finishing and the next one starting? Is there a `setTimeout` somewhere, or is there no async step between two tests?

static/resources/harness.js
```diff
@@ -144,59 +177,62 @@
...
         var value = eval(test.code);
-        if (value && typeof value === 'object' && 'result' in value) {
-          result.result = value.result;
-          if (value.message) {
-            result.message = value.message;
-          }
-        } else if (typeof value === 'boolean') {
-          result.result = value;
+
+        if ('Promise' in self && value instanceof Promise) {
```
foolip, line 204: Can we not instead detect a thenable here?

```js
if (typeof value === 'object' && typeof value.then === 'function') {
  value.then(process, process);
}
```

Writing that out, I also wonder what happens if a promise rejects with true or false? Seems like we should handle errors separately and make sure that always results in an inconclusive results (null) regardless of the rejection value.

static/resources/harness.js
vinyldarkscratch: Not currently, but this will allow us to do so.  This is useful for when we wish to add a custom message to test results.

static/resources/harness.js
vinyldarkscratch: You're right!  I'll update the code to make the rejection create an error object (which'll create a null result with the error as the message).

Regarding the detection method, I'm not sure what additional benefit that would grant us (at least for the time being), but I suppose that we could do it like that, yeah!

vinyldarkscratch: And I just realized the redundant use of `Promise.resolve()`, haha, whoops!

PR #740, opened 2020-11-13
Revert "Use process.env.npm_package_version vs. require('./package.json').version"

vinyldarkscratch:
Reverts foolip/mdn-bcd-collector#723


foolip:
Didn't it work on AppEngine?

vinyldarkscratch:
It did not, no -- the versions simply showed up blank.  From what I have determined, that environment variable is only set when running something with `npm start`.  I have no idea if it is also set for other package managers (such as `yarn` if we were to switch), and the original method was more recommended anyways.

PR #791, opened 2020-11-19
Compress results JSON with GZip compression

### LINKED ISSUE #790: GZip the results?
Should we consider using GZip compression to reduce the file size of the results?  Currently, each of our results files are about 4 MB, which is a lot of data being handled (~1.5 GB).  While I was reviewing https://github.com/mdn/browser-compat-data/pull/7398, I was experimenting with using zlib to compress and uncompress JSON, and I've seen a MAJOR file size decrease, with basically no performance drawbacks.

vinyldarkscratch:
This PR fixes #790.  To reduce file size, this uses GZip compression for the results JSON when saving them.  There are practically no performance drawbacks as a result.

### DIFF
--- a/github.js
+++ b/github.js
@@ -30,7 +30,7 @@ const github = (options) => {
     /* eslint-disable-next-line max-len */
     // like https://github.com/web-platform-tests/wpt.fyi/blob/26805a0122ea01076ac22c0a96313c1cf5cc30d6/results-processor/wptreport.py#L79
     const hash = crypto.createHash('sha1');
-    const digest = hash.update(buffer).digest('hex').substr(0, 10);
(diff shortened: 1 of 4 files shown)


foolip:
I'm not sure what the extent of this is, but we should not check compressed files into https://github.com/foolip/mdn-bcd-results, that will make that repository much larger over time, not smaller. That's because git repos are already compressed, and more importantly because git deduplicates the contents of similar files by basing one object on an existing similar one.

For web-platform-tests report files, which are also large JSON files, the difference in size between a collection of gzipped files and the uncompressed files in a git repository eventually gets to be something like a 100:1 ratio.

vinyldarkscratch:
By the git repo size, are you referring to the commit history and such?

PR #807, opened 2020-11-23
Remove formEncType and lowSrc custom IDL

foolip:
This is a capitalization typo:
https://github.com/mdn/browser-compat-data/pull/7402#discussion_r527274048


foolip:
@vinyldarkscratch did you also remove the custom IDL for the other capitalization typos we found? `onModule` is the latest I found, but that's not actually capitalization so would be missed by any script listing duplicates.

vinyldarkscratch:
Not yet, but I'm tracking this as #812!

PR #809, opened 2020-11-24
Fix test for MediaSource.isTypeSupported

vinyldarkscratch:


foolip:
Thanks for fixing this! I've filed https://github.com/foolip/mdn-bcd-collector/issues/825 as a follow-up since it might affect other things too.

PR #837, opened 2020-11-28
Remove miscapitalized entries

vinyldarkscratch:
This PR removes the duplicate, miscapitalized entries for the HTMLMarqueeElement interface.﻿


foolip:
Have all of these been renamed in BCD already?

vinyldarkscratch:
Yeah, these ones have been fixed in BCD.  Surprisingly enough, it looks like these were the only ones we were generating tests for.

Issue #431, opened 2020-09-08
Handle prefixes and alternative names


queengooborg:
> As discussed elsewhere, generating tests for all combinations of prefixes gets to be a bit excessive.
>
> It would be fine to split that change out from this larger change though, no need to fix everything at once.

_Originally posted by @foolip in https://github.com/foolip/mdn-bcd-collector/pull/416#discussion_r484442442_

Additional comment: https://github.com/foolip/mdn-bcd-collector/pull/416#discussion_r484452152

foolip:
To make sure we're not missing anything we can check against Confluence. `curl https://storage.googleapis.com/web-api-confluence-data-cache/latest/json/org.chromium.apis.web.GridRow.json | jq -r '.[] | .id' | grep -i webkit` should list all the WebKit stuff, for example.

vinyldarkscratch:
All prefixes are no longer being generated, so it's just the raw, unprefixed support now.  Unfortunately, I did take out the code that generates prefixed versions of all the properties based upon prefixes in BCD, so we'll need to add that back for both API and CSS.

foolip:
Our previous approach was to generate tests for paths that don't actually exist in BCD, like `css.properties.-webkit-display`. I think that we'll simplify interpretation of the results if we instead do something like this:

```js
bcd.addTest('api.Element.matches', function() {
  return 'webkitMatchesSelector' in Element.prototype;
}, /* new dictionary: */ {
  alternative_name: 'webkitMatchesSelector'
});
```

In the reports we would store this extra dictionary. It could also make sense to put the "method" used to test in there, and possibly the global scope tested.

Somewhat related to https://github.com/foolip/mdn-bcd-collector/issues/436.

gsnedders:
**Not a Contribution.**

Despite some CSS properties existing with `-apple-` (e.g., `-apple-pay-button-style`) and `-epub-caption-side` in WebKit/Blink, I think by and large we only need the more restricted set of `webkit`/`moz`/`ms` to cover everything (with varying casing for interfaces v. operations/attributes)?

This would be good to increase coverage where things are only supported prefixed (and to be able to show the initial version with support where it was initially prefixed), though would add a substantial number of new tests if runtime is a concern.

foolip:
[...]

PR #843, opened 2020-11-30
Add custom IDL for webkit-prefixed canvas APIs

foolip:


vinyldarkscratch:
Weren't we planning to generate the prefixes based upon data in BCD?  #431

foolip:
That was my intention initially, but simply adding the IDL also gets the job done. Being able to add some links explaining its origin is a nice bonus, although one that would be fine to give up if we do add support for generating these.

vinyldarkscratch:
Sounds good!  Let's merge this now then so we can get tests in the collector right away, and then later on we can hook it up as a generated version.

vinyldarkscratch (approved):

PR #947, opened 2021-02-04
IE 5.5 no longer supported

vinyldarkscratch:
Apparently IE 5.5 is no longer supported...ah well!  This is a quickie to update the compat table on the front.﻿

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -46,7 +46,7 @@
     </div>
     <div class="compat-table-separator"></div>
     <div class="compat-table-entry">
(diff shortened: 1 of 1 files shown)


foolip:
Do you mean it's not supported on BrowserStack, or that we accidentally made a change that broke the collector on IE 5.5?

vinyldarkscratch:
There's a change made on the collector that broke support on IE 5.5 -- I was working on replacing the IE ranged versions, and ran into an issue running the collector on IE 5.5.

foolip:
OK, let's just leave it broken and update the claimed support :)

PR #984, opened 2021-02-18
Drop support for building prefixed variants of tests

foolip:
Prefixed variants are now done via custom IDL and CSS. update-bcd.js
isn't able to map those back to the prefixed entries in BCD, then also
isn't able to do that for the mechanism now deleted.

This doesn't change the build tests.json at all because it was unused.

### DIFF
--- a/build.js
+++ b/build.js
@@ -151,38 +151,26 @@ const getCustomTestCSS = (name) => {
       compileCustomTest(customTests.css.properties[name]);
 };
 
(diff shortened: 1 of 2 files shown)


foolip:
> update-bcd.js isn't able to map those back to the prefixed entries in BCD

This was actually wrong, there is some support in update-bcd.js for prefixed entries. I don't know if it works though, since the results haven't had any prefixes in them.

PR #991, opened 2021-02-19
Validate the payloads sent to /api/results

foolip:
This is in preparation for exposing the results as a URL available to
another client (Selenium). It should be difficult to store arbitrary
data in the service to make it unattractive for abuse.

This also modifies the results format, notably dropping the code.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -218,7 +218,26 @@ POST /api/results
 ```
 
 The `Content-Type` should be `application/json` and the post body should be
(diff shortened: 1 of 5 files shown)


foolip:
@vinyldarkscratch note that there's some risk that this will break something. I'm especially unsure if there might be some message longer than 1000 posted from some browser. The longest message I can find in mdn-bcd-results has a length of 250, so I think it's OK, but something to keep a lookout for if you run the Selenium script and it's not working.

PR #997, opened 2021-02-22
Remove XML-style <br /> self-closing tags

foolip:
This was inconsistently applied, and not needed in HTML.

### DIFF
--- a/views/tests.ejs
+++ b/views/tests.ejs
@@ -11,23 +11,23 @@
   for (const [resourceId, resource] of Object.entries(resources)) {
     switch (resource.type) {
       case 'audio':
-      case 'video': 
+      case 'video':
         resourceCount += 1; %>
         <<%-resource.type%> id="resource-<%-resourceId%>">
           <% for (const src of resource.src) { %>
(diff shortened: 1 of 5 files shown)


views/tests.ejs
```diff
@@ -11,23 +11,23 @@
...
         <<%-resource.type%> id="resource-<%-resourceId%>">
           <% for (const src of resource.src) { %>
-            <source src="/resources/custom-tests<%-src%>" />
+            <source src="/resources/custom-tests<%-src%>">
           <% } %>
           <% if (resource.subtitles) {
             let i = 0;
             for (const sub of resource.subtitles) { %>
-            <track label="<%-sub.label%>" kind="subtitles" srclang="<%-sub.lang%>" src="/resources/custom-tests<%-sub.src%>" <% if (i == 0) { %>default<% } %> />
+            <track label="<%-sub.label%>" kind="subtitles" srclang="<%-sub.lang%>" src="/resources/custom-tests<%-sub.src%>" <% if (i == 0) { %>default<% } %>>
```
foolip, line 23: Confirmed manually that the `%>>` here doesn't break the template.

Issue #1000, opened 2021-02-22
update-bcd flag for only adding exact versions.


foolip:
From discussions with @gsnedders on IRC, it would be useful if the update-bcd script had a flag for only adding exact versions, no ranges like "≤14" or the "13> ≤14" values one can currently get if one has results for version 13 and 14 of a browser, but there's also a 13.1 version for which results are missing.

Such a flag would have one clear issue, namely that it would leave known-incorrect false and null values for things that are supported, but for which the exact version can't be determined.

gsnedders:
My issue on IRC was more along the lines of:

You can get this with results for 13.1 and 14.0 if there are tests that have only been run in 14.0 (e.g., if you have results for prod for 13.1, but results from staging for 14.0). I would, as a first pass, like to update results for things where we can give an exact version (i.e., where we ran the test in both browser versions).

foolip:
I see, so for those purposes the proposed flag would actually solve the problem, the difference is in why we don't have complete data in order to suggest an exact version.

foolip:
@gsnedders there's an implementation of this in https://github.com/foolip/mdn-bcd-collector/pull/1019 now if you'd like to try it out.

foolip:
I've split out just the documentation into https://github.com/foolip/mdn-bcd-collector/pull/1020 to document what currently works, on popular request.

PR #1019, opened 2021-02-24
Add --release argument (filter) for update-bcd.js

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/1000.


foolip:
Having tried to use this a bit I've always used first `--added` and then `--removed`, so I've just merged them into a single `--release` argument instead.

PR #1020, opened 2021-02-25
Document how to use the update-bcd script

foolip:
Part of https://github.com/foolip/mdn-bcd-collector/issues/1000.


foolip:
@Elchi3 would you mind trying these instructions and reporting back?

PR #1092, opened 2021-03-31
Update button/select/submit styling

vinyldarkscratch:
This PR is a minor update to the styling for buttons, select dropdowns, and submit buttons, performing the following:

- Applying the same pseudo-class styles as a regular button to submit buttons (hover, disabled)
- Ensuring disabled select dropdowns have the proper styling when not hovering

### DIFF
--- a/static/resources/style.css
+++ b/static/resources/style.css
@@ -58,12 +58,14 @@ input {
   margin-left: -2px;
 }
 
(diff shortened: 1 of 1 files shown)


foolip (approved):
Thanks, I've tried it locally and it (still) looks good!

Issue #12, opened 2019-03-27
CSS.supports can disagree with CSSOM feature detection


foolip:
Digging into https://github.com/mdn/browser-compat-data/pull/3667/files/5562af96cde102135a7e643e031f7f1fee71a3f7#diff-ca8acc47e6afb1f4ec0010894b1e7b03 I found that these two tests disagree in Edge:
- `CSS.supports("font-size-adjust", "inherit")` returns false
- `'fontSizeAdjust' in document.body.style` returns true

`CSS.supports("font-size-adjust", "0.5")` also returns false. Tested with https://software.hixie.ch/utilities/js/live-dom-viewer/saved/6808 in Edge 18.

I'd checked before that there were no disagreements, but might not have had these results then.

It will be necessary to have a way to resolve cases where two tests for the same feature lead to true and false.

foolip:
When resolving this issue we should check if https://github.com/foolip/mdn-bcd-collector/pull/416/files#r485522138 as been fixed. It's sort of in the same area of code.

gsnedders:
https://bugs.webkit.org/show_bug.cgi?id=222280 and https://bugs.webkit.org/show_bug.cgi?id=223141 will cause false-positives in WebKit-based browsers for runtime disabled properties when testing with `CSS.supports` (the only one I'm currently aware of is `overscroll-behavior`, but there may be others historically)

foolip:
Thanks @gsnedders, I hadn't noticed that, I hope we haven't updated data based on such false positives yet.

I think the useful next step here would try to depend only on `'fontSizeAdjust' in document.body.style` and see if that changes the results for any recent browsers. If not, we can just use that, and then use `CSS.supports` to check support for specific keywords in https://github.com/foolip/mdn-bcd-collector/issues/705.

foolip:
This issue affected https://mdn-bcd-collector.appspot.com/tests/css/properties/aspect-ratio in Safari 14.1, where `"aspectRatio" in document.body.style` returns false, but `CSS.supports("aspect-ratio", "inherit")` returns true.

@apple-web-evangelist FYI, there's a risk that other such false positives have occurred.

foolip:
[...]

PR #1132, opened 2021-04-29
Stop using CSS.supports for CSS property tests

foolip:
Instead generate both forms for the document.body.style tests where
appropriate, to deal with odd -apple- and -epub-prefixed properties.

Fixes https://github.com/foolip/mdn-bcd-collector/issues/12.

### DIFF
--- a/build.js
+++ b/build.js
@@ -158,9 +158,6 @@ const compileTestCode = (test) => {
   if (test.property.startsWith('constructor')) {
     return `bcd.testConstructor("${property}");`;
   }
-  if (test.owner === 'CSS.supports') {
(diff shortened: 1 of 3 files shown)


foolip:
Looking at the test results before and after this change, the false positives in Safari like `aspect-ratio` are fixed, no changes for Chrome, but Firefox results for "css.properties.-moz-context-properties" change from true to false.

That seems like a similar issue to the Safari false positives. I'll file a Gecko bug.

foolip:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1708384 and will merge this now, all the changes that I can see are improvements now.

PR #1133, opened 2021-04-29
Drop custom tests for -apple- and -internal-prefixed properties

foolip:
These shouldn't be web exposed and it will hopefully be fixed:
https://github.com/foolip/mdn-bcd-collector/issues/12#issuecomment-829084798

### DIFF
--- a/custom-css.json
+++ b/custom-css.json
@@ -1,9 +1,5 @@
 {
   "properties": {
-    "-apple-color-filter": {},
-    "-apple-pay-button-style": {},
-    "-apple-pay-button-type": {},
-    "-apple-trailing-word": {},
     "-epub-caption-side": {},
     "-epub-text-combine": {},
     "-epub-text-emphasis": {},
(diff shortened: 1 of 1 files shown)


custom-css.json
```diff
@@ -1,9 +1,5 @@
 {
   "properties": {
-    "-apple-color-filter": {},
```
gsnedders, line 3: This isn't internal, but should be disabled by default: https://bugs.webkit.org/show_bug.cgi?id=217802

gsnedders (commented):
FWIW, currently the properties with `"internal": true` are:

```
['-webkit-font-size-delta',
 '-webkit-marquee-direction',
 '-webkit-marquee-increment',
 '-webkit-marquee-repetition',
 '-webkit-marquee-speed',
 '-webkit-marquee-style',
 '-webkit-mask-repeat-x',
 '-webkit-mask-repeat-y',
 'background-repeat-x',
 'background-repeat-y']
 ```

foolip:
Testing https://mdn-bcd-collector.appspot.com/tests/css/properties in Safari 14.1 all of these are in fact still web exposed, so let's just keep the tests for them then. I'm skeptical that they should all be added to MDN/BCD, but that's a separate problem from generating the tests.

PR #1162, opened 2021-05-19
Update UA parser to handle old Android WebView versions

vinyldarkscratch:
This PR updates the UA parser script to handle pre-Chromium Android WebView versions.  Whilst setting up my virtual machines, I noticed that the built-in browser has its own separate version from the Android OS version (which BCD uses), somewhat like Safari iOS/iPadOS.  With this PR, the UA parser will use the Android OS version for pre-Chromium WebView versions.﻿

### DIFF
--- a/ua-parser.js
+++ b/ua-parser.js
@@ -36,6 +36,7 @@ const parseUA = (userAgent, browsers) => {
     case 'samsung_browser':
       data.browser.id = 'samsunginternet';
       break;
+    case 'android_browser':
     case 'chrome_webview':
       data.browser.id = 'webview';
       break;
@@ -48,7 +49,7 @@ const parseUA = (userAgent, browsers) => {
   }
 
   // https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#safari-for-ios-versioning
-  data.fullVersion = data.browser.id === 'safari_ios' ? ua.os.version : ua.browser.version;
+  data.fullVersion = (data.browser.id === 'safari_ios' || ua.browser.name === 'Android Browser') ? ua.os.version : ua.browser.version;
   data.version = getMajorMinorVersion(data.fullVersion);
 
   if (!(data.browser.id in browsers)) {


ua-parser.js
```diff
@@ -47,8 +48,19 @@ const parseUA = (userAgent, browsers) => {
     data.browser.name += ' Android';
   }
 
+  data.fullVersion = ua.browser.version;
   // https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#safari-for-ios-versioning
-  data.fullVersion = data.browser.id === 'safari_ios' ? ua.os.version : ua.browser.version;
+  if (data.browser.id === 'safari_ios') {
+    data.fullVersion = ua.os.version;
+  }
```
foolip, line 55: Can you go `else if` here to make it abundantly clear that `data.browser.id === 'safari_ios'` and  `ua.browser.name === 'Android Browser'` aren't expected to happen at the same time?

ua-parser.js
```diff
@@ -47,8 +48,19 @@ const parseUA = (userAgent, browsers) => {
...
+
+  if (ua.browser.name === 'Android Browser') {
+    if (compareVersions.compare(ua.os.version, '5.0', '<')) {
```
foolip, line 58: I'm pretty sure Chrome for Android 37 would have been available on Android 4.4.3, so is this the wrong check? How about instead checking if the engine is Chromium, with an assert that the version should be 37 or later?

foolip (commented):
Can you add some tests for this in unittest/unit/ua-parser.js?

ua-parser.js
vinyldarkscratch: The documentation that describes WebView indicates that Android 4.4.3's browser was based upon Chrome 33.  At the top, it has a statement that indicates that WebView 37 wasn't available before Android 5:

> In Android 5.0 (Lollipop), the WebView has moved to an APK so it can be updated seperately to the Android platform.

ua-parser.js
vinyldarkscratch: Done!

vinyldarkscratch:
Unittests added!

vinyldarkscratch:
(P.S. I think this PR gives a little idea of the current little project I'm working on, let's just say it holds true to my old-browser reputation.  >:D)

unittest/unit/ua-parser.js
```diff
@@ -117,7 +117,19 @@ describe('parseUA', () => {
...
-  it('WebView Android', () => {
+  it('WebView Android (Android Browser, 1.1)', () => {
+    assert.deepEqual(parseUA('Mozilla/5.0 (Linux; U; Android 1.1; en-us; generic) AppleWebKit/525.10+ (KHTML, like Gecko) Version/3.0.4 Mobile Safari/523.12.2', browsers), {browser: {id: 'webview_android', name: 'WebView Android'}, version: '1.1', fullVersion: '1.1', os: {name: 'Android', version: '1.1'}, inBcd: true});
+  });
+
+  it('WebView Android (Android Browser, 4.4.2, Chrome 30)', () => {
+    assert.deepEqual(parseUA('Mozilla/5.0 (Linux; Android 4.4.2; Android SDK built for x86 Build/KK) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/30.0.0.0 Mobile Safari/537.36', browsers), {browser: {id: 'webview_android', name: 'WebView Android'}, version: '4.4', fullVersion: '4.4.2', os: {name: 'Android', version: '4.4.2'}, inBcd: true});
+  });
+
+  it('WebView Android (Android Browser, 5.0.2, Chrome 39)', () => {
```
foolip, line 128:
```suggestion
  it('WebView Android (Android Browser, 5.0.2, Chrome 37)', () => {
```

Right?

foolip (approved):

Issue #1163, opened 2021-05-19
Start WebSocket server locally


gsnedders:
Currently we hardcode URLs in a few places:

```
gsnedders@gsnedders-milk mdn-bcd-collector % rg 'appspot\.com' --glob '!*.md' --glob '!unittest'
custom-tests.json
1360:      "__base": "if ('URL' in self) {var instance = new URL('https://mdn-bcd-collector.appspot.com')} else {var instance = new webkitURL('https://mdn-bcd-collector.appspot.com')}",
1455:      "__base": "var instance = new WebSocket('wss://mdn-bcd-collector.appspot.com');"

selenium.js
37:const host = `https://${testenv ? 'staging-dot-' : ''}mdn-bcd-collector.appspot.com`;
```

The most notable cases here are those in `custom-tests.json`, especially the `WebSocket` case. It would be nice to to use `location.host`, and ensure there's a WebSockets server running on the staging env and start one locally.

This is I _think_ the one case where we still talk to external hosts when running locally?

As an aside, it might be nice to replace the call of the URL constructor to something else, so we can more easily audit for hardcoding the staging/prod URL?

gsnedders:
> You also suggest running a WebSocket server, but I wonder if that's just to avoid console errors? Or are there instances we can only get with an actual server running? 

I don't know!  I just thought it did actually connect!

foolip:
I double checked and actually there's just the constructor, and nothing is done with the instance except feature detect members in generated tests. I also fixed another (harmless) external URL in https://github.com/foolip/mdn-bcd-collector/pull/1172.

Let's consider this fixed, running the tests should now hopefully not make any external connections except some webfonts and stuff for the styling of the page.

PR #1170, opened 2021-05-20
Avoid hardcoded appspot.com URLs in custom tests

foolip:
Part of https://github.com/foolip/mdn-bcd-collector/issues/1163

### DIFF
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -1357,7 +1357,7 @@
       "__base": "var instance = document.createTreeWalker(document);"
     },
     "URL": {
-      "__base": "if ('URL' in self) {var instance = new URL('https://mdn-bcd-collector.appspot.com')} else {var instance = new webkitURL('https://mdn-bcd-collector.appspot.com')}",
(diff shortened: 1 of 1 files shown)


custom-tests.json
```diff
@@ -1452,7 +1452,7 @@
       "__base": "<%api.EXT_disjoint_timer_query:ext%> var instance = ext.createQueryEXT();"
     },
     "WebSocket": {
-      "__base": "var instance = new WebSocket('wss://mdn-bcd-collector.appspot.com');"
+      "__base": "var instance = new WebSocket('wss://' + location.hostname);"
```
foolip, line 1455: @gsnedders I looked at `location.host` or `location.hostname` here, the difference being whether the port is included for cases like serving on http://localhost:8080. Since the scheme here is already hardcoded to wss:// my thinking is that including the port will sometimes be wrong. And by not included it there's a requirement to run this on port 443, effectively.

This doesn't matter until there's a WebSocket server to talk to, of course.

foolip:
GitHub Actions is not very well-behaved in this repo recently...

PR #1192, opened 2021-06-01
Get a DOMException instance from a thrown exception

foolip:
The exception type was supported long before its constructor in some
browsers, so this is necessary to get the right data.


foolip:
Ping @vinyldarkscratch

vinyldarkscratch (approved):
Clever!  Thanks for this change, it's certainly MUCH better than relying on the constructor!

foolip:
Thanks Vinyl!

PR #1194, opened 2021-06-02
Use navigator as the WorkerNavigator instance

foolip:
This avoid the potential problem with workers being supported but the
WorkerNavigator interface itself just not being exposed.

Also drop the window. prefix in some other tests, since it was
inconsistently used and skipping it makes Window+Worker code consistent.


vinyldarkscratch (approved):
Looks good to me, thanks for the update!

PR #1198, opened 2021-06-03
Fix custom test for ProcessingInstruction API

vinyldarkscratch:
This PR fixes an issue with the custom test for the `ProcessingInstruction` API.  When I initially wrote the custom test, I thought that it had to say `xml`, when in fact, it was the exact opposite.

### DIFF
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -838,7 +838,7 @@
       "__base": "var instance = navigator.plugins;"
     },
     "ProcessingInstruction": {
-      "__base": "var instance = document.createProcessingInstruction('bcd', \"version='1.0' encoding='UTF-8'\");"
+      "__base": "var instance = document.createProcessingInstruction('xml-stylesheet', 'type=\"text/xsl\" href=\"transform.xsl\"');"
     },
     "Range": {
       "__base": "var instance = document.createRange();"
--- a/custom-tests.json
+++ b/custom-tests.json
@@ -838,7 +838,7 @@
       "__base": "var instance = navigator.plugins;"
     },
     "ProcessingInstruction": {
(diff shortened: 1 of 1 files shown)


foolip:
Both `document.createProcessingInstruction('xml', "version='1.0' encoding='UTF-8'")` and `document.createProcessingInstruction('bcd', "version='1.0' encoding='UTF-8'")` work for me in Chrome. Is there some browser which throws an exception?

In any case, it wasn't a great example, because `<?xml version='1.0' encoding='UTF-8'?>` isn't a processing instruction. Maybe the example from https://developer.mozilla.org/en-US/docs/Web/API/Document/createProcessingInstruction#example would be better?

vinyldarkscratch:
Earlier versions of Safari and Firefox aren't too happy about using `xml` as the processing instruction's namespace.  This was causing some testing issues when working on https://github.com/mdn/browser-compat-data/pull/10724.

We could use the example direct from MDN web docs as well!  So long as we're not using `xml` for the first argument, we should be good.

custom-tests.json
```diff
@@ -838,7 +838,7 @@
       "__base": "var instance = navigator.plugins;"
     },
     "ProcessingInstruction": {
-      "__base": "var instance = document.createProcessingInstruction('xml', \"version='1.0' encoding='UTF-8'\");"
+      "__base": "var instance = document.createProcessingInstruction('bcd', \"version='1.0' encoding='UTF-8'\");"
```
foolip, line 841:
```suggestion
      "__base": "var instance = document.createProcessingInstruction('xml-stylesheet', 'type=\"text/xsl\" href=\"transform.xsl\"');"
```

This produces the second example from https://developer.mozilla.org/en-US/docs/Web/XML/XML_introduction#displaying_xml

foolip (approved):

PR #1202, opened 2021-06-03
Revert "Add reusable instances of Worker and SharedWorker"

foolip:
Reverts foolip/mdn-bcd-collector#1200


foolip:
After these changes I found that http://staging-dot-mdn-bcd-collector.appspot.com/tests/api/Worker/onerror doesn't work on Safari 5.1, the page just keeps reloading. I don't know what the problem is, maybe the worker script needs to actually do something. Whatever it is, it's not worth breaking testing on old Safaris.

@vinyldarkscratch FYI

PR #1210, opened 2021-06-07
Update custom test for ProcessingInstruction

vinyldarkscratch:
This updates the custom test for the ProcessingInstruction API, using the example from the MDN documentation, particularly to increase compatibility with Internet Explorer.  The original test wouldn't function in IE, throwing a "NotSupportedError".


foolip (approved):
Ah, I guess we need an XML document...

PR #1216, opened 2021-06-08
Simplify subtle.crypto custom test instance

foolip:


foolip:
Actually this isn't great, we should deal with `crypto.webkitSubtle` too.

vinyldarkscratch (approved):
This looks good to me, thanks!

foolip:
Follow-up in https://github.com/foolip/mdn-bcd-collector/issues/1217.

Issue #1057, opened 2021-03-16
missing api.HTMLAnchorElement.toString


gsnedders:
Somehow we aren't generating a test for the stringifier

foolip:
I've noticed this as well, and it also affects api.HTMLAreaElement.toString.

The problem is in `flattenMembers`, where attributes with `member.special === 'stringifier'` are not handled. It looks like there are more cases like this, but those `toString` entries aren't in BCD.

foolip:
Ah, and trying to fix this runs into https://github.com/foolip/mdn-bcd-collector/issues/841, because `"toString" in HTMLAnchorElement.prototype` is not a good-enough test for this.

foolip:
The generated test should rather be `HTMLAnchorElement.prototype.hasOwnProperty("toString")`, and I think this would be best handled in `buildIDLTests`.

PR #1226, opened 2021-06-10
Generate toString tests for all stringifiers, not just anonymous ones

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/1057.


foolip:
Fixed by https://github.com/foolip/mdn-bcd-collector/pull/1716.

PR #1227, opened 2021-06-10
Confluence list

foolip:


foolip:
I've been over the Confluence list enough times now that it's unlikely we've missed much of importance. Something like https://github.com/foolip/mdn-bcd-collector/pull/1225 should fit the role of discovering future missing things.

PR #1276, opened 2021-07-04
Selenium script updates

### LINKED ISSUE #1230: Selenium script leaves holes in the data that the update script gets tricked by
While working on demixing the `Body` mixin, I very nearly submitted bad data due to quirks of mdn-bcd-collector. The Selenium data collection script can leave holes in the data, and we had results for Firefox 10-35, 42 and 50-88, probably with some additional holes in those ranges.

The APIs were added in Firefox 39, but the update script generated &le;42, which is correct in a way, but since that is *also* often the result when attributes have been added and we have a null&rarr;true transition, it nearly fooled me.

The most important thing here is to fix the Selenium script to never leave holes in the data, it would be better to skip everything before the first hole I think.

vinyldarkscratch:
This PR performs a number of updates to the Selenium script to resolve some issues caused by some new custom tests, along with existing tests, due to permissions requirements.  This PR also performs a few other little updates to make the script a little nicer, including the following:

- Increase test timeout
- Turn the browser-specific ignores into arrays rather than strings
- Remove logic to remap Safari 6.1
- Allow mic, camera, geolocation, and notifications for Chrome, Edge and Firefox
- Simplify code to ignore the Safari x.0 versions on BrowserStack
- Increase verbosity of tests and add more comments
- Encode the commas in URLs for EdgeHTML versions of Edge (due to some odd issues)
- Add a command line argument to reverse the browser's version order in testing (oldest-to-newest)
- Catch results upload failures

This should fix #1230 with both the timeout increase and the permissions, since the holes are caused by timeouts (sometimes due to an on-screen prompt).

### DIFF
--- a/selenium.js
+++ b/selenium.js
@@ -153,6 +153,7 @@ const buildDriver = async (browser, version, os) => {
 
       capabilities.set(Capability.VERSION, version.split('.')[0]);
 
+      // Remap target OS for Safari x.0 vs. x.1 on SauceLabs
       if (service === 'saucelabs') {
         if (browser === 'safari') {
           capabilities.set('platform', getSafariOS(version));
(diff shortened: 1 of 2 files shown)


vinyldarkscratch:
It seems that Firefox is struggling at the moment.  I'm unsure of whether it's something to do with this script's updates, or if it's the recent custom tests we've added, but I feel it's probably this script.  Will need to look into it some more.

PR #1302, opened 2021-07-15
Generate tests for JavaScript builtins

foolip:
This is a manually generated list of tests to match as many existing BCD
paths as possible, starting from the output of find-missing.js.

A total of 517 tests are added. Almost all pass in Chrome + Firefox +
Safari, and those that fail have been vetted for reasonableness.

Fixes https://github.com/foolip/mdn-bcd-collector/issues/1056.


vinyldarkscratch (approved):
I'm totally down for adding tests for JavaScript to the collector, yeah!  This is looking good to me!

PR #1317, opened 2021-07-18
Upgrade lockfile version to 2

saschanaz:
It's required by npm 7.

Now that dependabot also supports it, I guess it's time to upgrade 👀

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8906 @@
 {
   "name": "mdn-bcd-collector",
   "version": "3.2.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "3.2.1",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@google-cloud/logging-winston": "4.1.0",
+        "@google-cloud/storage": "5.8.5",
+        "@mdi/font": "5.9.55",
+        "@mdn/browser-compat-data": "3.3.11",
+        "@octokit/rest": "18.6.7",
+        "compare-versions": "3.6.0",
+        "cookie-parser": "1.4.5",
+        "didyoumean": "1.2.2",
+        "ejs": "3.1.6",
+        "express": "4.17.1",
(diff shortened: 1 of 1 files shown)


foolip:
I'm not sure what version of npm is used on Google Cloud Build, but the fastest way to figure out is to merge this and see what happens :)

foolip:
The deploy seems to have worked, but I can't figure out from the logs which version of npm is used. That's a bit concerning to me, in the worst case it's npm 6 and this warning that I get locally applies:
> npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!

@saschanaz should that warning not be concerning?

foolip:
> It's required by npm 7.

Does this mean `npm install` doesn't work with an older lockfile? If npm 6 also doesn't fully support lockfile version 1, then how is one supposed to migrate without having full control over the version of npm used?

saschanaz:
> > It's required by npm 7.
> 
> Does this mean `npm install` doesn't work with an older lockfile? If npm 6 also doesn't fully support lockfile version 1, then how is one supposed to migrate without having full control over the version of npm used?

It means `npm install` by npm 7 will overwrite the lockfile by a new one. (Not sure whether it even reads version 1.) My vague memory says npm 5 showed a warning with version 2 without any actual error.

saschanaz:
The [release note](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile) says it's backwards compatible so it shouldn't be problematic on npm 6.

foolip:
OK, I guess I will ignore the warning until I upgrade to npm 7.

foolip:
Hmm, so when trying to do the new release I've run into a problem. I typically edit the "version" number in package.json and then run `npm i`, which updates the number in package-lock.json too. After this change, that will instead replace the whole package-lock.json with a new one with "lockfileVersion" 1. In other words, I guess npm 6 doesn't know how to update a newer package-lock.json.

I'm going to revert this and reconsider when minimum required node version is 16, which is not yet supported on AppEngine.

PR #1318, opened 2021-07-18
Cover inherited attributes

### LINKED ISSUE #1316: Cover inherited attributes
https://github.com/foolip/mdn-bcd-collector/blob/149912853621f7bec4d3c5afccb85c72e75779af/build.js#L259

Somehow the current code filters out inherited attributes, but since inherited attributes are virtually redefined, I think they shouldn't be filtered out.

```js
Object.getOwnPropertyDescriptor(DOMRect.prototype, "x") !== Object.getOwnPropertyDescriptor(DOMRectReadOnly.prototype, "x").
```

saschanaz:
Fixes #1316 (and my comments on mdn/browser-compat-data#11561 and mdn/browser-compat-data#11562)

### DIFF
--- a/build.js
+++ b/build.js
@@ -163,6 +163,9 @@ const compileTestCode = (test) => {
   if (test.property.startsWith('Symbol.')) {
     return `"Symbol" in self && "${test.property.replace('Symbol.', '')}" in Symbol && ${test.property} in ${test.owner}.prototype`;
   }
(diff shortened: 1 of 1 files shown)


foolip (approved):
Thanks! This will add 30 new api.DOM* tests, so I'll make a new release.

Issue #1281, opened 2021-07-05
Some test (or combination of) is freezing Firefox 34-41


queengooborg:
Some test is causing Firefox 34 through 41 to freeze up, causing crashes in the Selenium script.  There are other problems around that range as well in regards to microphone/camera permissions (going as far up as Firefox 52), so it's possible the mic/camera tests are causing more issues.

PR #1319, opened 2021-07-18
Ignore user media tests on Firefox 34-52

vinyldarkscratch:
This is a workaround for #1281 until we can figure out a better solution.  This PR disables the user media APIs for Firefox 34-52 to prevent the Selenium script freezing during these Firefox versions.


foolip (approved):

vinyldarkscratch:
Argh, looks like this didn't do the trick -- I'm still not getting results for these Firefox versions through Selenium...  I'll need to look into this more.

PR #1321, opened 2021-07-19
Revert "Upgrade lockfile version to 2"

foolip:
Reverts foolip/mdn-bcd-collector#1317


foolip:
Won't revert, seems to work.

foolip:
Actually, I will need to revert.

PR #1323, opened 2021-07-20
Fix variable names in custom tests to avoid "interface"

foolip:
This is apparently a reserved keyword in strict mode and is breaking
these tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mdn-bcd-collector",
-  "version": "3.2.1",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mdn-bcd-collector",
-  "version": "3.2.1",
+  "version": "3.2.2",
```
foolip, line 3: This is a drive-by fix, missed in the release.

PR #1332, opened 2021-07-22
Simplify getSupportMap to use bare values, not {result: ...}

foolip:
These objects only had a single property, making the code around it more
convoluted for no good, and harder to debug.

### DIFF
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -60,7 +60,7 @@ const getSupportMap = (report) => {
       } else if (result === false) {
         // This may yet be overruled by a later result (above).
         supported = false;
-      } else if (result == null) {
+      } else if (result === null) {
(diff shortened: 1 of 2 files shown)


update-bcd.js
```diff
@@ -48,29 +48,31 @@ const getSupportMap = (report) => {
...
+      if (result === true) {
+        // If any result is true, the flattened support should be true. There
+        // can be contradictory results with multiple exposure scopes, but here
+        // we treat support in any scope as support of the feature.
+        supported = true;
         break;
+      } else if (result === false) {
+        // This may yet be overruled by a later result (above).
+        supported = false;
+      } else if (result == null) {
```
foolip, line 63:
```suggestion
      } else if (result === null) {
```

Issue #1331, opened 2021-07-21
update-bcd.js doesn't handle different support across platforms correctly


foolip:
With the current 3.2.2 results in mdn-bcd-results, update-bcd doesn't do anything sensible for api.BluetoothDevice.ongattserverdisconnected, if trying to add that.

The `inferredStatements` returned from `inferSupportStatements` when using results from both macOS and Windows are:
```js
[
  { version_added: '57', version_removed: '58' },
  { version_added: '70' }
]
```

But if using only the macOS or Windows results one at a time, the `inferredStatements` are `[ { version_added: '56' } ]` for macOS and `[ { version_added: '70' } ]` for Windows. Putting those together should not result in what we actually see above.

foolip:
The problem is here:
https://github.com/foolip/mdn-bcd-collector/blob/8e8b117791e92903396391bd3022b9b8908a7532/update-bcd.js#L103-L120

We will just overwrite any previous result in `versionMap.set(version, supported)` rather than combining the results. We need code similar to in https://github.com/foolip/mdn-bcd-collector/pull/1332.

PR #1340, opened 2021-07-22
Combine results from different reports for the same browser version

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/1331.

### DIFF
--- a/update-bcd.js
+++ b/update-bcd.js
@@ -3,6 +3,7 @@
 
 'use strict';
 
+const assert = require('assert');
 const compareVersions = require('compare-versions');
 const fs = require('fs').promises;
 const klaw = require('klaw');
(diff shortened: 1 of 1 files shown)


foolip:
I've checked what the effect of this is in terms of what changes add-new-bcd will make. Only 5 BCD paths are affected:
- api.BluetoothDevice.ongattserverdisconnected
- api.BluetoothRemoteGATTCharacteristic.oncharacteristicvaluechanged
- api.Event.preventCapture
- api.Event.preventBubble
- api.TextTrackCue.getCueAsHTML

The `BluetoothDevice` cases are because of macOS / Windows differences for Chrome.

The `Event` cases are about Firefox, and have two reports for Firefox where the results were null (or missing?) in one and false in the other. I'll file an issue about this.

The `api.TextTrackCue.getCueAsHTML` case is because the Edge 83 for Windows results have null results here, with the error "threw TypeError: Cannot use 'in' operator to search for 'getCueAsHTML' in undefined". I don't know why, and it only happens in that single report. When this sort of thing happens, it'll now be silently smoothed over, which overall is for the good.

Later we might want to add some kind of auditing for APIs that should be marked as `partial_implementation` based on conflicting results.

PR #1355, opened 2021-08-01
Fix test for XPathResult

vinyldarkscratch:
Apparently, the resolver argument was required in earlier versions of Firefox (57 and below).  This PR corrects the test to work with these older Firefox versions.


foolip:
Thanks for fixing! I suspect we have more issues like this. We could look at tests that throw an exception to find more.

PR #1358, opened 2021-08-02
Convert custom-tests.json to YAML (for multiline formatting)

### LINKED ISSUE #1350: Custom tests format: switch to something that supports multi-line?
I've been thinking about this for a while, but I feel that we should consider switching the syntax format for our custom tests to something that would allow us to add multiple lines.  Unfortunately, JSON doesn't allow us to do that.

The reason why I'd like to do this is for the purposes of maintenance.  Commonly, whenever I'm modifying a custom test, I find myself copying the code from the JSON, formatting it with Prettier, making the modifications, re-minifying it, and putting it back into the JSON.  I feel it would be far more reasonable to skip all those steps, which would additionally increase legibility on our end.

Some options I can think of:
- XML -- its structure is very much similar to HTML, but it's hard to parse and we'd have to define much of the syntax
- YAML -- easy to read, easy to parse, and fairly straightforward conversion, but also not typically used in JS projects (this is my personal vote)
- JSON5 -- near identical to what we already have, and commonly used by Google, but also lacks syntax highlighting in many code editors
- Hjson -- similar to JSON5, and has syntax highlighting support, but not as popular
- JavaScript -- we could define the code directly in a .js file and use comments for the feature paths, which would allow us to utilize Prettier, but also hard to parse
- Our own syntax -- definitely a last-resort option 😛

vinyldarkscratch:
This PR converts the custom tests file from a JSON file to YAML, which fixes #1350.  This will make maintenance much easier, as the code will be formatted in the YAML file itself, complete with multi-line support.  I had chosen YAML due to its similar flexibility, support for comments, straightforward conversion, and multiline support without the use of escape characters.


vinyldarkscratch:
Urgh, the unittests are giving me a massive headache...  I'm not finding any way to mock the file without causing the imports to break.  :/

PR #1367, opened 2021-08-04
Add more custom tests

vinyldarkscratch:
This PR adds and updates various custom tests, such as enabling the BatteryManager custom tests (we've solved the browser freezing issue); adding custom tests for various IDB APIs, Clients, and FileReader; and adding support for callback-based tests (without needing to use promises).

### DIFF
--- a/build.js
+++ b/build.js
@@ -44,13 +44,14 @@ const compileCustomTest = (code, format = true) => {
         return `throw 'Test is malformed: ${match} is an invalid reference';`;
       }
       let importcode = compileCustomTest(customTests.api[name].__base, false);
+      const callback = importcode.includes('callback');
 
       importcode = importcode
           .replace(/var (instance|promise)/g, `var ${instancevar}`)
           .replace(/callback\(/g, `${instancevar}(`)
           .replace(/promise\.then/g, `${instancevar}.then`)
           .replace(/(instance|promise) = /g, `${instancevar} = `);
-      if (instancevar !== 'instance' && instancevar !== 'promise') {
+      if (!(['instance', 'promise'].includes(instancevar) || callback)) {
         importcode += ` if (!${instancevar}) {return false;}`;
       }
       return importcode;
--- a/build.js
+++ b/build.js
@@ -688,7 +688,6 @@ const buildJS = (customJS) => {
 const copyResources = async () => {
   const resources = [
     ['json3/lib/json3.min.js', 'resources'],
-    ['promise-polyfill/dist/polyfill.min.js', 'resources', 'promise.min.js'],
     ['chai/chai.js', 'unittest'],
     ['mocha/mocha.css', 'unittest'],
     ['mocha/mocha.js', 'unittest'],
--- a/build.js
+++ b/build.js
@@ -684,6 +684,7 @@ const buildJS = (customJS) => {
 const copyResources = async () => {
   const resources = [
     ['json3/lib/json3.min.js', 'resources'],
+    ['promise-polyfill/dist/polyfill.min.js', 'resources', 'promise.min.js'],
     ['chai/chai.js', 'unittest'],
     ['mocha/mocha.css', 'unittest'],
     ['mocha/mocha.js', 'unittest'],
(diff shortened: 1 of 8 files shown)


build.js
```diff
@@ -44,12 +44,14 @@ const compileCustomTest = (code, format = true) => {
         return `throw 'Test is malformed: ${match} is an invalid reference';`;
       }
       let importcode = compileCustomTest(customTests.api[name].__base, false);
+      const callback = importcode.includes('callback');
 
       importcode = importcode
           .replace(/var (instance|promise)/g, `var ${instancevar}`)
+          .replace(/callback\(/g, `${instancevar}(`)
```
foolip, line 51: This looks strange. If the test code has `callback(something)`, does it make sense to replace it with `instance(something)`? Maybe it does, but then the variable name `instancevar` is no longer accurate and rather misleading.

custom-tests.yaml
```diff
@@ -214,8 +214,13 @@ api:
...
         return ctx.decodeAudioData.length === 1;
-  # BatteryManager: # This promise seems to remain pending when running all tests
-  #   __base: var promise = navigator.getBattery();
+  BatteryManager:
+    __base: >-
+      if (!('getBattery' in navigator)) {
+        return false;
+      };
+      var promise = navigator.getBattery();
+      promise.then(function() {});
```
foolip, line 223: Why is this `promise.then(function() {})` needed? It doesn't silence any rejected promise and seems like a no-op.

custom-tests.yaml
```diff
@@ -933,6 +942,23 @@ api:
   HTMLVideoElement:
     __base: var instance = document.createElement('video');
     __test: return bcd.testObjectName(instance, 'HTMLVideoElement');
+  IDBDatabase:
+    __base: >-
+      <%api.IDBOpenDBRequest:req%>
+      req.onsuccess = function() {
+        callback(req.result);
```
foolip, line 949: In the context of these tests, `callback()` + `fail()` looks like an odd pairing. Why not `success()` or `result()` + `fail()` instead? Both success and failure uses callbacks, after all, so both could be called "callback" :)

static/resources/harness.js
```diff
@@ -242,29 +243,34 @@
...
-            function(fail) {
-              processTestResult(new Error(fail), data, i, callback);
-            });
-        value['catch'](
-            function(err) {
-              processTestResult(err, data, i, callback);
-            });
-      } else {
-        processTestResult(value, data, i, callback);
+        value.then(success, fail);
```
foolip, line 267: Aha, so it looks like the promise support could be built on top of the callbacks. Good!

static/resources/harness.js
```diff
@@ -242,29 +243,34 @@
...
+        value['catch'](fail);
```
foolip, line 268: Is both this line and the previous needed? It seems like we should either have just `value.then(success, fail)` or `value.then(success)` + `value['catch'](fail)`. Is this due to one of those forms not being supported somewhere early on? If so a comment about it would be great.

views/index.ejs
```diff
@@ -60,10 +60,11 @@
 <div id="known-caveats" class="paper">
   <p>Known caveats:</p>
   <ul>
-    <li>Detection for features under prefixes are not yet supported</li>
+    <li>Detection for features under prefixes are not fully supported</li>
     <li>Detection for features under alternative names are not yet supported</li>
     <li>In older browsers, some attributes aren't initialized in the API prototypes, resulting in false negatives (see <a href="https://github.com/foolip/mdn-bcd-collector/issues/406">#406</a>)</li>
     <li>Some APIs are accessed through other APIs and aren't exposed on their own; custom tests have not been written for them all</li>
+    <li>The test page uses a JSON polyfill for core functionality; tests for <code>javascript.builtins.JSON</code> should not be trusted</li>
```
foolip, line 67: This didn't change in this PR, but seems serious. Is there any issue filed about this? We should delete the tests if they can't be trusted.

custom-tests.yaml
vinyldarkscratch: It's an oddball statement, but there seems to be a bug in early Firefox versions where the promises from the navigator will daisy-chain `.then` statements even when using separate calls to `getBattery()` or `getUserMedia()`.  This seems to resolve the issue, though I couldn't say how or why.

PR #1409, opened 2021-08-16
Fix test for RTCPeerConnection

vinyldarkscratch:
- Fix test for RTCPeerConnection for older browsers
- Use bcd.testConstructor to check if the constructor is one

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -1500,7 +1500,7 @@ api:
       if (!constructor) {
         return false;
       }
-      var instance = new constructor();
+      var instance = new constructor({iceServers: []});
   Screen:
     __base: var instance = window.screen;
   ScriptProcessorNode:
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -1500,6 +1500,10 @@ api:
       if (!constructor) {
         return false;
       }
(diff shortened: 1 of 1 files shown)


custom-tests.yaml
```diff
@@ -1500,7 +1500,11 @@ api:
       if (!constructor) {
         return false;
       }
-      var instance = new constructor();
+      var cnstResult = bcd.testConstructor(constructor);
```
foolip, line 1503: @vinyldarkscratch is this bit really needed, would just adding the `{iceServers: []}` argument not have sufficed? `RTCPeerConnection` has always been a thing you construct, it will never have thrown an exception with the kind of message that `bcd.testConstructor` looks for.

foolip: I've sent https://github.com/foolip/mdn-bcd-collector/pull/1424.

vinyldarkscratch: It does, actually -- in early versions of Firefox when Web RTC was enabled by a flag, the `RTCPeerConnection` constructor _was_ defined, but attempting to construct something with it returns an error saying it "is not a constructor", thus resulting in null values.

foolip: I see, can you add a comment to the test?

PR #1422, opened 2021-08-20
Revert "Updates for event custom tests"

foolip:
Reverts foolip/mdn-bcd-collector#1405

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -122,14 +122,7 @@ api:
       try {
         instance = new AnimationEvent('animationend');
       } catch(e) {
-        try {
-          instance = document.createEvent('AnimationEvent');
-        } catch(e) {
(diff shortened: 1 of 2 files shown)


foolip:
@vinyldarkscratch I created this revert for discussion so I don't forget. I'm not sure that I trust that a lack of support for an event in `document.createEvent()` reliably means the event interface isn't supported.

At some point maybe 5 years ago Chromium was updated to only support a fixed list of interfaces via `document.createEvent()`. At that time, and perhaps even recently, there could be events that don't have constructors and that aren't on the list of interfaces in `document.createEvent()`.

If we can't create an instance of the event interface, I would rather treat it as inconclusive and require us to look at it manually.

foolip:
> so I don't forget

This didn't work, or not fast enough at least :)

@queengooborg what do you think about the problem here?

PR #1425, opened 2021-08-20
Revert "Make testConstructor able to test constructor objects"

foolip:
Reverts foolip/mdn-bcd-collector#1408


foolip:
I think this code will be unused after https://github.com/foolip/mdn-bcd-collector/pull/1424, in which case we can remove it.

PR #1426, opened 2021-08-20
Add comment explaining the RTCPeerConnection constructor test

vinyldarkscratch:
Addresses review comments in https://github.com/foolip/mdn-bcd-collector/pull/1409#discussion_r692946179.  Closes #1424, closes #1425.﻿

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -1506,7 +1506,7 @@ api:
         return false;
       }
 
-      // Earlier versions of Firefox expose the "RTCPeerConnection" variable
+      // Earlier versions of Firefox expose mozRTCPeerConnection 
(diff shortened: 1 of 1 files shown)


custom-tests.yaml
```diff
@@ -1505,10 +1505,15 @@ api:
       if (!constructor) {
         return false;
       }
+
+      // Earlier versions of Firefox expose the "RTCPeerConnection" variable
```
foolip, line 1509:
```suggestion
      // Earlier versions of Firefox expose mozRTCPeerConnection 
```

Was it `mozRTCPeerConnection` at that time? Trying to avoid the word "variable" here since it's rather a property on the global object.

foolip (approved):

PR #1433, opened 2021-08-23
Add custom test for DOMTokenList.toggle.force_parameter

vinyldarkscratch:


custom-tests.yaml
```diff
@@ -580,6 +580,14 @@ api:
         elm.className = ' foo bar foo ';
         elm.classList.remove('bar');
         return elm.className === 'foo';
+      toggle.force_parameter: >-
+        var elm = document.createElement('b');
+        elm.className = 'foo bar';
+        if (!('classList' in elm && 'toggle' in elm.classList)) {
```
foolip, line 586: This could be simpler: `if (!elm.classList || !elm.classList.toggle)`

PR #1436, opened 2021-08-26
Add documentation for how to review changes from the collector

foolip:


vinyldarkscratch:
I just checked this documentation out, and it looks really awesome, great work!  It just about covers everything important for reviews.  😄

There's only one remaining section I think we should add, which may involve a little more in-depth work.  What about when a test changes in a new collector version, but the PR was submitted using older results?  (I'm usually proactive about this, but I know I've sometimes slipped, usually me forgetting to bump the collector version in my PR-creating script!)

foolip:
> What about when a test changes in a new collector version, but the PR was submitted using older results?

What should a reviewer do in this situation? I have just ignored the possibility of this, thinking that eventually we'll vet all differences between the collector results and BCD, so as long as the changes seem plausible, I haven't worried about which collector version the results are from.

foolip:
@ddbeck I've linked to a bunch of examples now, do you want to give this another look?

ddbeck:
Looks great! This is really helpful and I'd love to see this linked from new collector-based PRs in the future. 👍

foolip:
@queengooborg can you review and update your collector description templates to link this?

queengooborg (approved):
Thanks for your hard work writing this up, @foolip!  Let's get it merged!  (I'll update my PR creation script to link to this section right away!

PR #1444, opened 2021-08-27
Add custom tests for MediaStreamEvent + MediaStreamTrackEvent

foolip:


vinyldarkscratch:
We have support for callback-based tests now, but we've yet to write a custom test that includes both callback- and promise-based tests, with the ability to switch between the two...  We'll have to look into that.  Should we merge this as-is and open an issue regarding the callback-based method of obtaining a `MediaDevices` interface?

PR #1464, opened 2021-09-06
Revert "Create custom test for RTCRtpReceiver"

foolip:
Reverts foolip/mdn-bcd-collector#1459

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -1529,10 +1529,6 @@ api:
       }
 
       var instance = new constructor({iceServers: []});
-  RTCRtpReceiver:
-    __base: >-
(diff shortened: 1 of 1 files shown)


foolip:
@queengooborg this test won't work because `getReceivers()` returns an array of `RTCRtpReceiver` objects, and it could be an empty array.

queengooborg:
Do you have any suggestions on how to resolve this?  Could we maybe try creating a connection before trying to get them?

foolip:
Given how new `RTCRtpReceiver` is, are you sure we need a custom test at all? Missing interface objects is mostly a thing of the past.

PR #1469, opened 2021-09-08
Fix custom test for RTCPeerConnection

queengooborg:
This PR fixes the custom test for the RTCPeerConnection API.  Apparently in these earlier versions, it was actually that the non-prefixed version was unintentionally exposed and wasn't a valid constructor.  The test now attempts to use the prefixed variant first to resolve this issue.﻿


foolip:
Curious :)

A comment to explain the order would be helpful, otherwise this is the sort of thing I'd fix in a drive-by edit for consistency, not realizing the order matters.

PR #1474, opened 2021-09-09
Use document.fonts or self.fonts as FontFaceSet instance

foolip:
This needs a custom test because the interface has
[LegacyNoInterfaceObject] in Chromium.


foolip:
@queengooborg this works, but I'd rather generate different code for window and worker tests. Is that supported somehow?

queengooborg:
We don't have a way to write separate custom tests depending on the scope; your solution is pretty much what I've done for other custom tests.

This is looking good to me by the way!

foolip:
OK, let's go with this then.

PR #1475, opened 2021-09-09
Convert from CJS to ESM

### LINKED ISSUE #1443: Switch to ESM?
This issue is inspired by https://github.com/mdn/browser-compat-data/pull/12161.  It might be beneficial for us to switch our repository to ESM, since BCD's internals seem to be heading in that direction?  Luckily, ESM can import a CommonJS module (but not the other way around, sadly), so we don't have to wait for BCD to make a switch first.

queengooborg:
This PR converts our project from CJS to ESM to keep up with various tools, BCD included, making the switch.  This transition included a number of internal changes for our code infrastructure, particularly involving imports, exports, and the `__dirname` variable (AKA file paths), but nothing else has changed much.  Resolves #1443.

Package changes:
- Replaces `proxyquire` embedded import switching (I couldn't get a require mock to work for ESM)
- Replaces the deprecated `babel-eslint` with `@babel/eslint-parser`
- Replaces `nyc` with `c8`, which is actively developed and maintained and has all the same features with important ESM bugfixes
- Adds `es-main` to replace `require.main === module` (I would just use a one-liner but this package is small, and the one-liner check doesn't work on Windows...thanks, Microsoft)

Successfully tested components:
- Main app
  - Launching app (macOS, Windows, and Linux)
  - Running a test
  - Exporting the test to a downloadable file
- Build
- Selenium
- add-new-bcd
- update-bcd
- Unittests


queengooborg:
Whew, the ESM conversion is finally all done!  Would you like to take a look at this, @foolip?

add-new-bcd.js
```diff
@@ -123,7 +122,7 @@ const main = async () => {
   console.log('Done!');
 };
 
-main().catch((e) => {
-  console.error(e);
-  process.exit(1);
-});
+/* istanbul ignore if */
+if (esMain(import.meta)) {
+  main();
```
foolip, line 127: Don't we still need to catch promise rejection here?

app.js
```diff
@@ -225,37 +229,32 @@ app.use((req, res) => {
...
-              default: 8443
-            })
-            .option('port', {
-              describe: 'HTTP port',
-              type: 'number',
-              default: process.env.PORT ? +process.env.PORT : 8080
-            });
-      }
+if (esMain(import.meta)) {
+  const {argv} = yargs(hideBin(process.argv)).command(
```
foolip, line 234: Is this change involving `hideBin` required here, or can it be done separately? Doesn't just `yargs.command(...)` continue to work here?

app.js
```diff
@@ -14,38 +14,42 @@
...
-const logger = require('./logger');
-const {parseResults} = require('./results');
-const storage = require('./storage').getStorage();
-const {parseUA} = require('./ua-parser');
+const storage = getStorage();
 
 const appVersion =
   process.env.GAE_VERSION === 'production' ?
-    require('./package.json').version :
+    JSON.parse(await fs.readFile('./package.json')).version :
```
foolip, line 42: Since `fs` is fs-extra, can you use https://github.com/jprichardson/node-fs-extra/blob/master/docs/readJson.md here and elsewhere to reduce repetition?

build.js
```diff
@@ -14,28 +14,32 @@
...
 );
-const customCSS = require('./custom-css.json');
-const customIDL = require('./custom-idl');
-const customJS = require('./custom-js.json');
 
-const generatedDir = path.join(__dirname, 'generated');
+const customCSS = JSON.parse(await fs.readFile('./custom-css.json'));
+const customJS = JSON.parse(await fs.readFile('./custom-js.json'));
+
+const generatedDir = fileURLToPath(new URL('./generated', import.meta.url));
```
foolip, line 42: Is this really the most straightforward way to get a path relative to the current file with ESM? Going via URLs seems roundabout to me...

foolip (commented):
I haven't finished reviewing, submitting the comments I have so far.

app.js
queengooborg: This was taken directly from the Yargs example.  All of the examples, including CommonJS examples, have this new structure so I think there's a good reason for it.

build.js
queengooborg: This is what the BCD conversion does for replacing `__dirname` as well.  I've tried other methods that don't use URLs but I've had no luck.

add-new-bcd.js
queengooborg: One nice feature ESM offers is top-level async, so we don't have to worry about writing a `catch()` block.  Admittedly though, I wasn't using it correctly, as one still needs to `await` the function.  Fixed!

foolip: Oh, I'm glad I asked then :)

app.js
foolip: Fair enough, I guess it does something useful :)

build.js
foolip: https://stackoverflow.com/questions/46745014/alternative-for-dirname-in-node-when-using-the-experimental-modules-flag points to this solution. I don't think it's very pretty, but alright then :)

package.json
```diff
@@ -68,15 +74,13 @@
...
     "klaw": "3.0.0",
     "listr2": "3.11.1",
     "minimatch": "3.0.4",
     "mocha": "9.1.1",
     "node-fetch": "2.6.1",
     "nodemon": "2.0.12",
-    "nyc": "15.1.0",
     "open-cli": "7.0.1",
     "prettier": "2.4.0",
     "proxyquire": "2.1.3",
```
foolip, line 86: Can we drop this dependency now?

app.js
```diff
@@ -14,38 +14,44 @@
...
+import {getStorage} from './storage.js';
+import {parseUA} from './ua-parser.js';
+import Tests from './tests.js';
+
+const storage = getStorage();
 
 const appVersion =
   process.env.GAE_VERSION === 'production' ?
-    require('./package.json').version :
+    (await fs.readJson('./package.json'))?.version :
```
foolip, line 42: Not sure why `?.` was added, the only way `await fs.readJson('./package.json')` could be null is if the file contains just the string "null", which is a surprising error we shouldn't mask. Suggest reverting:

```suggestion
    (await fs.readJson('./package.json')).version :
```

build.js
```diff
@@ -14,28 +14,32 @@
...
-const WebIDL2 = require('webidl2');
-const YAML = require('yaml');
+import css from '@webref/css';
+import esMain from 'es-main';
+import fs from 'fs-extra';
+import idl from '@webref/idl';
+import path from 'path';
+import prettier from 'prettier';
+import {fileURLToPath} from 'url';
+import {validate} from 'webidl2';
```
foolip, line 24:
```suggestion
import * as WebIDL2 from 'webidl2';
```

To avoid a bunch of changes that aren't strictly required. I also think it's clearer in context to have `WebIDL2.validate()` or `WebIDL2.parse()`, not the bare functions which have very generic names.

build.js
```diff
@@ -786,7 +790,9 @@ const copyResources = async () => {
     ['@mdi/font/fonts/materialdesignicons-webfont.woff2', 'fonts']
   ];
   for (const [srcInModules, destInGenerated, newFilename] of resources) {
-    const src = require.resolve(srcInModules);
+    const src = fileURLToPath(
+      new URL(`./node_modules/${srcInModules}`, import.meta.url)
+    );
```
foolip, line 795:
```suggestion
    const src = `./node_modules/${srcInModules}`;
```

build.js
```diff
@@ -806,11 +812,16 @@ const build = async (customIDL, customCSS) => {
   const JSTests = buildJS(customJS);
   const tests = Object.assign({}, IDLTests, CSSTests, JSTests);
 
-  await fs.writeJson(path.join(__dirname, 'tests.json'), tests);
+  await fs.writeJson(new URL('./tests.json', import.meta.url), tests);
```
foolip, line 815:
```suggestion
  await fs.writeJson('./tests.json', tests);
```

build.js
foolip: Actually, elsewhere with `fs.readFile` and `fs.readJson` we already assume that relative paths will work. So let's just assume the same here to avoid the extra stuff:

```suggestion
const generatedDir ='./generated';
```

custom-idl/index.js
```diff
@@ -14,14 +14,15 @@
 
 'use strict';
 
-const fs = require('fs');
-const path = require('path');
-const WebIDL2 = require('webidl2');
+import fs from 'fs-extra';
+import path from 'path';
+import {fileURLToPath} from 'url';
+import {parse} from 'webidl2';
```
foolip, line 20:
```suggestion
import * as WebIDL2 from 'webidl2';
```

custom-idl/index.js
```diff
@@ -14,14 +14,15 @@
...
 
 // Load text (UTF-8) files from a directory and return an object mapping each
 // name (sans extension) to the parsed result of that text.
-const parseIDL = () => {
-  const files = fs.readdirSync(__dirname);
+const parseIDL = async () => {
+  const files = await fs.readdir(fileURLToPath(new URL('.', import.meta.url)));
```
foolip, line 25: To reduce the ugly stuff with URLs, can you do `const dirname = fileURLToPath(import.meta.url)` once and then use `await fs.readdir(dir)` here?

custom-idl/index.js
```diff
@@ -30,10 +31,13 @@ const parseIDL = () => {
       continue;
     }
     const name = path.parse(file).name;
-    const text = fs.readFileSync(path.join(__dirname, file), 'utf8');
-    results[name] = WebIDL2.parse(text);
+    const text = await fs.readFile(
+      fileURLToPath(new URL(`./${file}`, import.meta.url)),
+      'utf8'
+    );
```
foolip, line 37:
```suggestion
    const text = await fs.readFile(path.join(dirname, file), 'utf8');
```

exporter.js
```diff
@@ -74,15 +76,20 @@ const createBody = (meta) => {
   );
 };
 
-const exportAsPR = async (report, token) => {
-  const octokit = new Octokit({auth: `token ${token}`});
+const exportAsPR = async (report, token, octokit) => {
```
foolip, line 79: Rather than passing in both `token` and `octokit`, how about we just always pass in `octokit`? That would mean creating an `Octokit` instance in the /export endpoint, and makes the test and production code more similar. This would be a case of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), and is how one often ends up having to structure things for testability in less dynamic languages.

Some of the following code would also be simplified.

find-missing-features.js
```diff
@@ -1,8 +1,21 @@
...
+import esMain from 'es-main';
+import fs from 'fs-extra';
+import yargs from 'yargs';
+import {hideBin} from 'yargs/helpers';
+
 const BCD_DIR = process.env.BCD_DIR || `../browser-compat-data`;
-const bcd = require(BCD_DIR);
-const tests = require('./tests.json');
+const {default: bcd} = await import(
+  process.env.NODE_ENV === 'test' ?
```
foolip, line 10: How about adding `bcd` and `tests` arguments to `main()` and putting the production bits of this in the `if (esMain(import.meta)` branch?

scripts.js
```diff
@@ -24,8 +26,8 @@ const exec = (cmd, env) => {
 
 const prepare = () => {
   // Copy secrets.sample.json to secrets.json if needed
-  const secretsPath = path.join(__dirname, 'secrets.json');
-  const secretsSamplePath = path.join(__dirname, 'secrets.sample.json');
+  const secretsPath = new URL('./secrets.json', import.meta.url);
```
foolip, line 29: Will this really work without the `fileURLToPath` bit? Does `fs.existsSync` accept a file:// URL?

selenium.js
```diff
@@ -12,26 +12,32 @@
...
+import bcd from '@mdn/browser-compat-data';
+const bcdBrowsers = bcd.browsers;
+import compareVersions from 'compare-versions';
+import fetch from 'node-fetch';
+import esMain from 'es-main';
+import fs from 'fs-extra';
+import path from 'path';
+import chalk from 'chalk';
+import listr2 from 'listr2';
+const {Listr} = listr2;
```
foolip, line 33: Can't this be simplified into one line somehow? Has listr2 done something unusual with their exports?

selenium.js
```diff
@@ -517,40 +523,41 @@ const runAll = async (limitBrowsers, oses, nonConcurrent, reverse) => {
...
+          alias: 'r',
+          type: 'boolean',
+          nargs: 0
+        });
+    }
   );
 
   if (
-    runAll(argv.browser, argv.os, argv.nonConcurrent, argv.reverse) === false
+    (await runAll(argv.browser, argv.os, argv.nonConcurrent, argv.reverse)) ===
```
foolip, line 559: Since a rejected promise is now automatically turned into a non-zero exit code, can we instead throw exceptions in `runAll()` for the cases where we return `false` now and just `await runAll(argv.browser, argv.os, argv.nonConcurrent, argv.reverse)` here instead?

storage.js
```diff
@@ -92,13 +91,13 @@ class MemoryStorage {
 
   async saveFile(filename, data) {
     assert(!filename.includes('..'));
-    const p = path.join(__dirname, 'download', filename);
+    const p = fileURLToPath(new URL(`./download/${filename}`, import.meta.url));
```
foolip, line 94: Can you use `fileURLToPath` once to set a `dirname` constant and simplify here and below?

unittest/app/app.js
```diff
@@ -14,20 +14,24 @@
...
 
-const {app, version} = require('../../app');
+import fs from "fs-extra";
+import {fileURLToPath} from "url";
+
+import {app, version} from "../../app.js";
 const agent = chai.request.agent(app);
 
-const tests = Object.entries(require('../../tests.json'));
+const tests = Object.entries(JSON.parse(await fs.readFile(
```
foolip, line 27: Can you extract out a `dirname` variable here and also use `fs.readJson()`?

unittest/puppeteer/harness.js
```diff
@@ -14,15 +14,17 @@
...
 
-const fs = require('fs');
-const path = require('path');
+import {app} from "../../app.js";
 
-const puppeteer = require('puppeteer');
-const pti = require('puppeteer-to-istanbul');
-
-const {app} = require('../../app');
+const pkg = JSON.parse(await fs.readFile(
```
foolip, line 25: Same `dirname` and `fs.readJson` comment here.

unittest/unit/build.js
```diff
@@ -14,16 +14,14 @@
...
-const chaiSubset = require('chai-subset');
+import chai, {assert, expect} from "chai";
+import chaiSubset from "chai-subset";
 chai.use(chaiSubset);
-const assert = chai.assert;
-const expect = chai.expect;
 
-const WebIDL2 = require('webidl2');
-const sinon = require('sinon');
+import {parse} from "webidl2";
```
foolip, line 21:
```suggestion
import * as WebIDL2 from "webidl2";
```

unittest/unit/exporter.js
```diff
@@ -59,16 +60,8 @@ const REPORTS = [
   }
 ];
 
-describe('GitHub export', () => {
+describe('GitHub export', async () => {
```
foolip, line 63:
```suggestion
describe('GitHub export', () => {
```

Last I tried, async functions for `describe` doesn't do what one would hope, the framework doesn't wait for the promise to resolve, so it really does nothing. Hopefully not needed?

Although... it is used below. Maybe that's actually broken too?

unittest/unit/find-missing-features.js
```diff
@@ -14,74 +14,30 @@
...
-  '../../find-missing-features',
-  {
-    './tests.json': tests,
-    '../browser-compat-data': bcd
-  }
-);
+import bcd from './bcd.test.js';
 
-describe('find-missing', () => {
+describe('find-missing', async () => {
```
foolip, line 25: An addition of `async` without any added `await` is suspect:

```suggestion
describe('find-missing', () => {
```

unittest/unit/update-bcd.js
```diff
@@ -399,7 +296,7 @@ const reports = [
   }
 ];
 
-describe('BCD updater', () => {
+describe('BCD updater', async () => {
```
foolip, line 299:
```suggestion
describe('BCD updater', () => {
```

update-bcd.js
```diff
@@ -3,19 +3,34 @@
...
+import minimatch from 'minimatch';
+const {Minimatch} = minimatch;
+import path from 'path';
+import yargs from 'yargs';
+import {hideBin} from 'yargs/helpers';
+
+import logger from './logger.js';
+import {parseUA} from './ua-parser.js';
+
+const overrides = (await fs.readJson(
```
foolip, line 20: Can you arrange for this to be passed in instead, similar to previous comments?

foolip (changes requested):

unittest/unit/exporter.js
queengooborg: This was probably a leftover from when I had been trying to configure the `proxyquire` alternative.

scripts.js
queengooborg: It seems that file URLs are indeed supported by `fs`, see https://nodejs.org/dist/latest-v14.x/docs/api/fs.html#fs_url_object_support.  From what I've also researched on that page, we can also use relative paths, which will resolve relative to `process.cwd()`, which we only change when preparing Puppeteer.

unittest/app/app.js
queengooborg: It looks like `fs-extra`, like NodeJS' built-in `fs`, accepts URL objects, so we can simplify this pretty easily!

build.js
queengooborg: I've found that it's actually better if we continue to use URL objects.  `fs` supports URL objects natively (and in turn, so does `fs-extra`), so we won't need to use `fileURLToPath` everywhere.  However, if we pass a string as a relative file path, `.` is resolved to the current working directory, which we cannot always trust to be the root of the project.

foolip: Oh, cool, that does make things nicer. Docs at https://nodejs.org/api/fs.html#fs_file_url_paths.

selenium.js
```diff
@@ -513,13 +515,7 @@ const runAll = async (limitBrowsers, oses, nonConcurrent, reverse) => {
...
   });
 
-  try {
-    await taskrun.run({testenv});
-    return true;
-  } catch (e) {
-    console.error(e);
-    return false;
-  }
+  await taskrun.run({testenv});
```
foolip, line 518: Cool, just letting the errors propagate simplified this too 👍

foolip (changes requested):

exporter.js
```diff
@@ -76,13 +75,9 @@ const createBody = (meta) => {
...
 
-const exportAsPR = async (report, token, octokit) => {
-  if (!token) {
-    return null;
-  }
-
+const exportAsPR = async (report, octokit) => {
   if (!octokit) {
-    octokit = new Octokit({auth: `token ${token}`});
+    return null;
```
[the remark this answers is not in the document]
foolip: Final remaining issue here :)

foolip (approved):

foolip:
Thanks @queengooborg for this huge PR! I gave up quickly on my last attempt, thanks for sticking with it :)

queengooborg:
Whoo, super exciting to get this merged!  🎉

Issue #1478, opened 2021-09-10
FetchEvent tests fail silenty


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/FetchEvent never finishes, tested in Chrome, Firefox and Safari.

There's this error in the console:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at Object.runTests (harness.js:378)
    at serviceworker.js:34
```

foolip:
Here's the code:
https://github.com/foolip/mdn-bcd-collector/blob/34516e5a123261f9c6b373dd2fe959a4a5fd87a7/static/resources/harness.js#L378-L380

foolip:
`tests` is undefined, because `pending.Worker` is undefined here:
https://github.com/foolip/mdn-bcd-collector/blob/34516e5a123261f9c6b373dd2fe959a4a5fd87a7/static/resources/harness.js#L530-L536

foolip:
I'm working on a fix.

PR #1479, opened 2021-09-10
Send the right tests to SharedWorker and ServiceWorker

foolip:
The tests for Worker were always sent here, which is right often enough
that nobody noticed until now. However, when there are no Worker tests,
it breaks.

This doesn't change the generated tests, but it will change which tests
are actually run, and so might have a big impact on the results.

Fixes https://github.com/foolip/mdn-bcd-collector/issues/1478.


queengooborg (approved):
This was most likely a copy-paste error from my end, thanks for catching this!

PR #1481, opened 2021-09-10
Add required request parameter to FetchEvent constructor

foolip:
Even with this change the test doesn't work as far back as FetchEvent is
probably supported, but at least it now works for current browsers.


queengooborg (approved):
LGTM, thanks!  :+1:

PR #1484, opened 2021-09-11
Automatically tag new releases

queengooborg:
This PR updates the `push` workflow to automatically tag new releases as needed.


foolip (approved):

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/1490 to test this.

queengooborg:
Looks like this action requires some sort of "token" parameter, which makes sense.  I'll see if I can test this on a test repo and figure it out!

PR #1497, opened 2021-09-17
Add custom test for HTMLOptionsCollection

queengooborg:


foolip (approved):
Creating multiple options is probably not necessary, but also is more realistic than 0 or 1 option elements, so 👍

PR #1499, opened 2021-09-17
Add a privacy notice to homepage

queengooborg:
Just a nice little thing to include on the homepage!﻿

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -73,7 +73,7 @@
 <div id="privacy-notice" class="paper">
   <h2><span class="mdi mdi-alert"></span> Privacy Notice</h2>
   <p>This website is designed to test as many browser features as possible. This includes features revealing personal information, such as microphone, camera, geolocation, etc. <strong>No personal information is stored locally or transmitted to any servers.</strong></p>
-  <p>We also use cookies to track your session.</p>
+  <p>We also use cookies to track your session and link the test results to your session.</p>
 </div>
 
 <script>
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -58,7 +58,7 @@
 </div>
 
 <div id="known-caveats" class="paper">
-  <p>Known caveats:</p>
+  <h2>Known Caveats</h2>
   <ul>
     <li>Detection for features under prefixes are not fully supported</li>
     <li>Detection for features under alternative names are not yet supported</li>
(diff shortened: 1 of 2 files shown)


views/index.ejs
```diff
@@ -70,6 +70,12 @@
   <p class="error-notice">If you find any errors in feature detection, please <a href="https://github.com/foolip/mdn-bcd-collector/issues/new"><span class="mdi mdi-alert-circle-outline"></span> file an issue</a> in the GitHub repository.</p>
 </div>
 
+<div id="privacy-notice" class="paper">
+  <h2><span class="mdi mdi-alert"></span> Privacy Notice</h2>
+  <p>This website is designed to test as many browser features as possible. This includes interfaces containing personal information, such as the microphone, camera, geolocation, etc. All of this information is kept local and is not sent to a server nor included in results.</p>
```
foolip, line 75:
```suggestion
  <p>This website is designed to test as many browser features as possible. This includes features revealing personal information, such as microphone, camera, geolocation, etc. All of this information is kept local and is not sent to a server nor included in results.</p>
```

views/index.ejs
```diff
@@ -70,6 +70,12 @@
...
+  <p>Other than the browser's user agent string, which is used to identify the browser and browser version tests are run on, all data is anonymous.</p>
```
foolip, line 76: I'm not sure about this bit, the UA string isn't personally identifying. Maybe just say that we store the UA string and why?

foolip (commented):
Great initiative, thanks!

views/index.ejs
queengooborg: I recalled hearing that a browser's UA string is considered personal information, but reading the data again I now see that it's only considered so when paired with an IP address.  Will fix this right away!

views/index.ejs
```diff
@@ -70,6 +70,12 @@
   <p class="error-notice">If you find any errors in feature detection, please <a href="https://github.com/foolip/mdn-bcd-collector/issues/new"><span class="mdi mdi-alert-circle-outline"></span> file an issue</a> in the GitHub repository.</p>
 </div>
 
+<div id="privacy-notice" class="paper">
+  <h2><span class="mdi mdi-alert"></span> Privacy Notice</h2>
+  <p>This website is designed to test as many browser features as possible. This includes features revealing personal information, such as microphone, camera, geolocation, etc. <strong>No personal information is stored locally or transmitted to any servers.</strong></p>
+  <p>We also use cookies to track your session and link the test results to your session.</p>
```
foolip, line 76:
```suggestion
  <p>Cookies are used for session tracking, so that test results can be stored on the server and then optionally exported. Test results are deleted after 15 days.</p>
```

foolip (approved):

PR #1501, opened 2021-09-17
Fix import order in update-bcd script

queengooborg:
This fixes an issue where the call to `yargs` is looking for the `browsers` variable, but it isn't initialized yet.


queengooborg:
I was about to merge this to fix this critical bug right away, but I forgot we're requiring review on PRs now...

foolip (approved):

PR #1418, opened 2021-08-19
Updates color contrast ratios to ensure they are meeting WCAG 2.0 AA

tannerdolby:
Fixes #1410 

This should resolve the remaining issues from the Lighthouse audit. I think everything looks good now that we are WCAG AA compliant for color contrast. Feel free to pull into this branch and have a look or view the short demo below.

Updated colors:
- button, select, input - Bg: `#3D853F` / Foreground: `#FFF` - Ratio: 4.54: 1
- button:hover, select:hover, input[type=submit]:hover - Bg: `#0F8A13` / Foreground: `#FFF` - Ratio: 4.5:1
- input::placeholder - Bg: `#FEFEFE` / Foreground: `#3D853F` - Ratio: 4.5:1

https://user-images.githubusercontent.com/48612525/130120045-324286f4-605f-48e4-8bfc-7ceca9801a86.mov

PR #1502, opened 2021-09-18
Sass it up!

queengooborg:
This PR converts our stylesheet into an SCSS file, with functions to create the gradients and handle dark vs. light themes properly.  This is intended to help us change our colors easier if we ever desire to in the future (an example of this is #1418).


package.json
```diff
@@ -86,6 +86,7 @@
     "prettier": "2.4.1",
     "puppeteer": "10.2.0",
     "puppeteer-to-istanbul": "1.4.0",
+    "sass": "^1.41.1",
```
foolip, line 89:
```suggestion
    "sass": "1.41.1",
```

foolip (approved):
There's a merge conflict, but looks good.

PR #1508, opened 2021-09-21
Remove insufficient tests for document.createElement() options

foolip:
These tests will return true even if the options are completely ignored.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -528,13 +528,6 @@ api:
       }
   Document:
     __base: var instance = document;
-    __additional:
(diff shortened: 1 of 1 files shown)


foolip:
@queengooborg can you check if BCD was updated based on these tests? If so, support likely goes back much further than it really should.

queengooborg (approved):
I had set the version numbers using commit history in mdn/browser-compat-data#8483 rather than the tests, probably since the tests were added after my BCD PR.  I don't see any updates to those lines that were from collector data.

PR #1509, opened 2021-09-21
Add custom test for api.Element.scroll.options_behavior_parameter

foolip:


foolip:
This is just a single test to show what it would look like. To test all of the combinations, it would be 18 tests. 6 different scroll methods, and 3 parameters for each.

PR #1513, opened 2021-09-21
Create release preparation script

queengooborg:
This PR creates a brand new release preparation script that bumps the version and updates the changelog (soon to be created in a separate PR), and then creates a PR using the GitHub CLI.  Depends on #1514.


release.js
```diff
@@ -0,0 +1,121 @@
...
+  for (const f of ['./package.json', './package-lock.json']) {
+    const filepath = new URL(f, import.meta.url);
+    const data = await fs.readJson(filepath);
+    data.version = newVersion;
+    await fs.writeJson(filepath, data, {spaces: 2});
+  }
+};
+
+const getChanges = () => {
+  const changes = exec(`git log --pretty=reference v${currentVersion}..main`)
```
foolip, line 56: I don't have a local `head` branch (can't push to it anyway) so this didn't work for me. This will:

```suggestion
  const changes = exec(`git log --pretty=reference v${currentVersion}..HEAD`)
```

release.js
```diff
@@ -0,0 +1,121 @@
...
+    getChanges() +
+    '\n\n' +
+    changelog.substring(idx, changelog.length);
+  await fs.writeFile(filepath, newChangelog, 'utf8');
+};
+
+const prepareBranch = (newVersion) => {
+  const branch = `release-${newVersion}`;
+
+  exec('git checkout main');
```
foolip, line 79: This step fails for me, the script just made local changes and checkout fails with "Your local changes to the following files would be overwritten by checkout".

I think the script should first ensure that there are no local changes to begin with (before making the bump changes), then check out the new release branch, and only after that list changes to make the changelog.

release.js
```diff
@@ -0,0 +1,121 @@
...
+  exec(`git branch ${branch}`);
+  exec(`git checkout ${branch}`);
+  exec('git stage package.json package-lock.json CHANGELOG.md');
+  exec(
+    `git commit -m "Release ${newVersion}" -m "" -m "Release v${newVersion}, generated by \\\`release.js\\\`."`
+  );
+  exec('git checkout main');
```
foolip, line 86: Just staying on the new branch would be better I think. `main` wasn't necessarily the original branch, and staying on the release branch makes it faster to check what happened.

release.js
```diff
@@ -0,0 +1,121 @@
...
+    changelog.substring(idx, changelog.length);
+  await fs.writeFile(filepath, newChangelog, 'utf8');
+};
+
+const prepareBranch = (newVersion) => {
+  const branch = `release-${newVersion}`;
+
+  exec('git checkout main');
+  exec(`git branch ${branch}`);
+  exec(`git checkout ${branch}`);
```
foolip, line 81:
```suggestion
  exec(`git checkout -b ${branch} origin/main`);
```

scripts.js
```diff
@@ -18,10 +18,12 @@ import fs from 'fs';
 import yargs from 'yargs';
 import {hideBin} from 'yargs/helpers';
 
-const exec = (cmd, env) => {
+export const exec = (cmd, env, pipe = true) => {
   env = {...process.env, ...env};
-  console.log(`> ${cmd}`);
-  return childProcess.execSync(cmd, {env, stdio: 'inherit'});
+  if (!pipe) {
+    console.log(`> ${cmd}`);
```
foolip, line 24: Is this leftover debugging? It looks a lot like the format that `npm run bla` prints, but why print anything?

foolip (changes requested):

scripts.js
queengooborg: This function is used by `npm run prepare` and `npm test` as well, which is why there's this console log.

release.js
```diff
@@ -112,6 +115,7 @@ const main = async () => {
     console.log(chalk`{yellow Release cancelled by user}`);
   }
 
+  exec('git checkout main');
```
foolip: I don't want a local `main` branch, the only thing it can do is get out of sync with origin.

```suggestion
  exec('git checkout origin/main');
```

queengooborg: I actually removed this checkout line completely, and any checkout of a local `main` branch!

foolip:
@queengooborg can you rebase / resolve conflicts?

release.js
```diff
@@ -0,0 +1,201 @@
...
+    exec('gh --version');
+  } catch (e) {
+    console.error(
+      chalk`{red This script depends on the {bold GitHub CLI}. Please {bold install} the CLI using the following instructions:} {blue https://cli.github.com/}`
+    );
+    return false;
+  }
+
+  console.log(chalk`{blue Fetching from remote...}`);
+  await gitRepo.fetchAll();
```
foolip, line 42: This step fails for me:
```
[Error: authentication required but no callback set] {
  errno: -1,
  errorFunction: 'Remote.fetch'
}
```

foolip (approved):
I'm going to merge this since it's so big, let's iterate on some details after landing.

PR #1514, opened 2021-09-21
Create a changelog

### LINKED ISSUE #1446: Create a changelog
Now that we have more reviewers for PRs created from this project, I think it would be very helpful to include a changelog, which we can link on the collector's page.  I feel that this should involve creating a script that handles the new release process, kind of like the scripts in BCD.

queengooborg:
This PR creates a changelog for all of the past releases since v1.0, using a quick Python script:

```py
import subprocess
from distutils.version import StrictVersion

v_call = subprocess.run(['git', 'tag'], stdout=subprocess.PIPE)
versions = [v[1:] for v in v_call.stdout.decode().split('\n')[:-1]]
versions.sort(key=StrictVersion)

print('# mdn-bcd-collector Changelog\n')

for i in reversed(range(len(versions)-1)):
  call = subprocess.run(['git', 'log', '--pretty=reference', f'v{versions[i]}..v{versions[i+1]}'], stdout=subprocess.PIPE)
  all_changes = [c[9:-1].replace('<', '&lt;').replace('>', '&gt;').replace('*', '\\*') for c in call.stdout.decode().split('\n')[1:-1]]
  changes = filter(lambda c: not c.startswith('Bump '), all_changes)

  print(f'## v{versions[i+1]}\n\n- '+'\n- '.join(changes)+'\n')

print('## v1.0.0\n\nInitial release!')

```

Fixes #1446.


CHANGELOG.md
```diff
@@ -0,0 +1,990 @@
+# mdn-bcd-collector Changelog
+
+## v3.3.1
+
+- (Automatically tag new releases (#1484), 2021-09-15)
+- (Add RTCSessionDescription test; fix RTCDTMFToneChangeEvent (#1488), 2021-09-15)
+- (Bump @google-cloud/storage from 5.14.1 to 5.14.2, 2021-09-14)
```
foolip, line 7: Can you filter out all of the dependency upgrades, or perhaps summarize them as a single "Upgraded dependencies" item at the end of each release? (That would avoid the problem of an empty changelog if only deps changes.)

foolip (changes requested):

CHANGELOG.md
queengooborg: I think it makes sense to state updates to the webref and BCD packages, but the rest is not useful to collector consumers.  I'll strip out all non-webref and non-BCD updates!

CHANGELOG.md
```diff
@@ -0,0 +1,558 @@
+# mdn-bcd-collector Changelog
+
+## v3.3.1
+
+- (Automatically tag new releases (#1484), 2021-09-15)
```
foolip, line 5: Hmm, why is each line in (parentheses)?

CHANGELOG.md
```diff
@@ -0,0 +1,558 @@
...
+- (Bump @webref/css from 2.0.1 to 2.0.2, 2021-07-01)
+- (Bump @mdn/browser-compat-data from 3.3.9 to 3.3.10, 2021-07-01)
+- (Bump @webref/idl from 2.2.3 to 2.2.4, 2021-06-25)
+- (Bump @webref/css from 2.0.0 to 2.0.1, 2021-06-24)
+- (Bump @mdn/browser-compat-data from 3.3.8 to 3.3.9, 2021-06-24)
+
+## v3.1.8
+
+- (Add a --path argument to filter BCD paths by wildcards (#1245), 2021-06-21)
+- (Use <b> as a HTMLElement instance (not HTMLUnknownElement) (#1253), 2021-06-21)
```
foolip, line 184: The `<b>` here makes the rest of the changelog bold. Is there a Markdown escaping function we can use?

CHANGELOG.md
```diff
@@ -0,0 +1,756 @@
+# mdn-bcd-collector Changelog
+
+## v3.3.1
+
+- Automatically tag new releases (#1484), 2021-09-15
+- Add RTCSessionDescription test; fix RTCDTMFToneChangeEvent (#1488), 2021-09-15
+- Bump @google-cloud/storage from 5.14.1 to 5.14.2, 2021-09-14
```
foolip, line 7: Still a lot of bumping going on here :)

queengooborg: I set it up so the bumps get moved down to the bottom of the entries, but don't have any special magic for combining these bumps together.  I could manually combine these if you'd like?

foolip: Let's filter out all the bump commits, and if there were any filtered out, add a "Upgraded dependencies" line to the changelog.

queengooborg: I opted to omit any mention of a dependency upgrade, as quite literally every release seems to feature a dependency upgrade.  All dependency bump entries have now been removed!

foolip (approved):

PR #1528, opened 2021-09-25
Add custom tests for HTMLFormControlsCollection and RadioNodeList

queengooborg:


custom-tests.yaml
```diff
@@ -879,6 +879,17 @@ api:
   HTMLFontElement:
     __base: var instance = document.createElement('font');
     __test: return bcd.testObjectName(instance, 'HTMLFontElement');
+  HTMLFormControlsCollection:
+    __base: >-
+      <%api.HTMLFormElement:form%>
+      for (var i = 0; i < 3; i++) {
+        var input = document.createElement('input');
+        input.type = "radio";
```
foolip, line 887: The properties aren't necessary to set for this test, right?

It's used for the next test, but better then to not reuse the code I think, so that each test is close to minimal for what is being tested.

foolip (changes requested):

custom-tests.yaml
```diff
@@ -1517,6 +1515,10 @@ api:
   RadioNodeList:
     __base: >-
       <%api.HTMLFormControlsCollection:controls%>
+      for (var i = 0; i < controls.length; i++) {
```
foolip, line 1518: This will result in two loops in the generated test. I was thinking rather starting from `<%api.HTMLFormElement:form%>` and just duplicating some of the code. Simplicity of the generated tests is more important than code reuse internally.

foolip (changes requested):

custom-tests.yaml
```diff
@@ -1503,6 +1512,18 @@ api:
...
+      <%api.HTMLFormElement:form%>
+      for (var i = 0; i < 3; i++) {
+        var input = document.createElement('input');
+        input.value = i;
+        input.type = "radio";
+        input.name = "radio";
+        form.appendChild(input);
+      }
+      var controls = form.elements;
+      var instance = controls.namedItem('radio');
```
foolip, line 1526: Even simpler:

```suggestion
      var instance = form.elements.namedItem('radio');
```

foolip (changes requested):

foolip (approved):

PR #1539, opened 2021-10-01
Use NodeGit's Revwalk to generate the changelog

foolip:
https://www.nodegit.org/api/revwalk/

Also remove dates from the changelog since they aren't included in the
summary by default, and aren't really useful to add either.


queengooborg (approved):
I'm all for simplifying our dependencies, this is looking good to me!  👍

PR #1558, opened 2021-10-12
Map WindowOrWorkerGlobalScope members to _globals folder

queengooborg:
This PR maps the members of the `WindowOrWorkerGlobalScope` to the new feature paths, handling their move to the new `_globals` folder that BCD had introduced.  The code is messy and very much special-casing the mixin...but it works.  :P

### DIFF
--- a/build.js
+++ b/build.js
@@ -300,6 +300,10 @@ const flattenIDL = (specIDLs, customIDLs) => {
   // mix in the mixins
   for (const dfn of ast) {
     if (dfn.type === 'includes') {
+      if (dfn.includes === 'WindowOrWorkerGlobalScope') {
+        // WindowOrWorkerGlobalScope is mapped differently in BCD
+        continue;
+      }
       const mixin = ast.find(
         (it) =>
           !it.partial &&
@@ -327,7 +331,10 @@ const flattenIDL = (specIDLs, customIDLs) => {
 
   // drop includes and mixins
   ast = ast.filter(
-    (dfn) => dfn.type !== 'includes' && dfn.type !== 'interface mixin'
+    (dfn) =>
+      (dfn.type !== 'includes' && dfn.type !== 'interface mixin') ||
+      // WindowOrWorkerGlobalScope is mapped differently in BCD
+      dfn.name === 'WindowOrWorkerGlobalScope'
   );
 
   return ast;
@@ -539,7 +546,12 @@ const buildIDLTests = (ast) => {
   const tests = {};
 
   const interfaces = ast.filter((dfn) => {
-    return dfn.type === 'interface' || dfn.type === 'namespace';
+    return (
+      dfn.type === 'interface' ||
+      dfn.type === 'namespace' ||
+      // WindowOrWorkerGlobalScope is mapped differently in BCD
+      dfn.name === 'WindowOrWorkerGlobalScope'
+    );
   });
   interfaces.sort((a, b) => a.name.localeCompare(b.name));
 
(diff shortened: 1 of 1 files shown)


build.js
```diff
@@ -327,7 +331,10 @@ const flattenIDL = (specIDLs, customIDLs) => {
 
   // drop includes and mixins
```
foolip, line 332:
```suggestion
  // drop includes and mixins, except WindowOrWorkerGlobalScope
```

build.js
```diff
@@ -539,7 +546,12 @@ const buildIDLTests = (ast) => {
   const tests = {};
 
   const interfaces = ast.filter((dfn) => {
-    return dfn.type === 'interface' || dfn.type === 'namespace';
+    return (
+      dfn.type === 'interface' ||
+      dfn.type === 'namespace' ||
+      // WindowOrWorkerGlobalScope is mapped differently in BCD
```
foolip, line 552: Rather than adding the `WindowOrWorkerGlobalScope` definition to `interfaces` here, how about doing `const globals = ast.find((dfn) => dfn.name === 'WindowOrWorkerGlobalScope')` and dealing with `globals` before or after `interfaces`? That should avoid a lot of special casing inside the `for (const iface of interfaces)` loop, and there really isn't a lot of code in common to share. The generated test would just be `'fetch' in self` and similar.

foolip (changes requested):

build.js
```diff
@@ -565,70 +646,15 @@ const buildIDLTests = (ast) => {
...
-
-      tests[`api.${iface.name}.${member.name}`] = compileTest({
-        raw: {
-          code: expr
-        },
-        exposure: Array.from(exposureSet),
-        resources: resources
-      });
-      handledMemberNames.add(member.name);
+    const memberTests = buildIDLChildrenTests(
```
foolip, line 649:
```suggestion
    const memberTests = buildIDLMemberTests(
```

build.js
```diff
@@ -565,70 +646,15 @@ const buildIDLTests = (ast) => {
...
+      members,
+      iface,
+      exposureSet,
+      isGlobal,
+      resources
+    );
+    for (const [k, v] of Object.entries(memberTests)) {
+      tests[`api.${iface.name}.${k}`] = v;
```
foolip, line 657: If the BCD path is instead constructed in `buildIDLMemberTests`, then this can probably be removed, and custom tests can use the BCD paths too.

unittest/unit/build.js
```diff
@@ -509,6 +509,45 @@ describe('build', () => {
...
+      const {ast, globals} = flattenIDL(specIDLs, customIDLs);
+      assert.lengthOf(ast, 3);
+      assert.lengthOf(globals, 1);
+
+      assert.equal(ast[0].name, 'Window');
+      assert.lengthOf(ast[0].members, 1);
+      assert.containSubset(ast[0].members[0], {
+        type: 'attribute',
+        name: 'imadumdum'
+      });
```
foolip, line 538: This should already be tested by some other test. I don't think testing the stuff that remains in `ast` in depth here is necessary.

unittest/unit/build.js
```diff
@@ -1135,6 +1174,37 @@ describe('build', () => {
         }
       });
     });
+
+    it('Globals', () => {
+      const ast = WebIDL2.parse(
+        `[Exposed=Window]
+           interface Window {
+             readonly attribute boolean imadumdum;
```
foolip, line 1182: Maybe drop this to make the test simpler. Every extra detail that's tested is more tests to update if we change something about how tests for `Window` members are generated.

queengooborg: I wanted to include some other interface to reflect a typical use case with both `ast` and `globals` populated.  I'll rename this dummy Window API though to make sure we won't run into any conflicts here!

PR #1572, opened 2021-10-15
Ignore Safari backported releases

queengooborg:
This PR updates the UA parser to ignore known backported releases of Safari, rather than remapping them to the versions they backport (due to OS integration differences causing some features to report as false) or worse, to their prior versions.  This is practically the opposite of #1244.

### DIFF
--- a/ua-parser.js
+++ b/ua-parser.js
@@ -74,6 +74,15 @@ const parseUA = (userAgent, browsers) => {
   const versions = Object.keys(browsers[data.browser.id].releases);
   versions.sort(compareVersions);
 
+  // Certain Safari versions are backports of newer versions, but contain less
+  // features, particularly ones involving OS integration. We will ignore these
+  // versions instead of remapping them to other versions to avoid confusion.
+  if (data.browser.id === 'safari') {
+    if (['4.1', '6.1', '6.2', '7.1'].includes(data.version)) {
+      return data;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


ua-parser.js
```diff
@@ -74,6 +74,15 @@ const parseUA = (userAgent, browsers) => {
   const versions = Object.keys(browsers[data.browser.id].releases);
   versions.sort(compareVersions);
 
+  // Certain Safari versions are backports of newer versions, but contain less
+  // features, particularly ones involving OS integration. We will ignore these
+  // versions instead of remapping them to other versions to avoid confusion.
+  if (data.browser.id === 'safari') {
+    if (['4.1', '6.1', '6.2', '7.1'].includes(data.version)) {
```
foolip, line 81: Combine into the previous `if` with `&&`?

ua-parser.js
```diff
@@ -74,6 +74,15 @@ const parseUA = (userAgent, browsers) => {
   const versions = Object.keys(browsers[data.browser.id].releases);
   versions.sort(compareVersions);
 
+  // Certain Safari versions are backports of newer versions, but contain less
+  // features, particularly ones involving OS integration. We will ignore these
```
foolip, line 78: It's not ignored, something is returned. I guess `inBcd` being false is the important thing?

foolip: Can you update the comment to reflect this?

foolip (approved):

Issue #1049, opened 2021-03-10
Detected Browser on /export page can show the wrong browser


foolip:
It seems like this code isn't actually running on every request:
https://github.com/foolip/mdn-bcd-collector/blob/d9722f79748a45010f54446cdbee269690e6990c/app.js#L76-L80

This manifests for me as the detected browser on /export sometimes being "Chrome 88" when I'm collecting results from other browsers. I'm on Chrome 89 now, but presumably the last browser to trigger that code was Chrome 88.

Fixing this is probably a matter of just not using `app.locals.browser` and instead extracting the UA string in the same code that invokes the template.

PR #1578, opened 2021-10-15
Keep UA parse results specific to the request, not the app

queengooborg:
This PR fixes the bug addressed in #1049.  After reading the documentation for Express, it turns out that I was misusing `app.locals`.  `app.locals` is for the lifetime of the _server,_ which is fitting for variables like the app's version.  For per-request variables, however, we want to use `res.locals` instead.﻿

### DIFF
--- a/app.js
+++ b/app.js
@@ -87,9 +87,11 @@ app.use(express.json({limit: '32mb'}));
 app.use(express.static('static'));
 app.use(express.static('generated'));
 
+app.locals.appVersion = appVersion;
+
+// Get user agent
 app.use((req, res, next) => {
-  app.locals.appVersion = appVersion;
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -87,9 +87,11 @@ app.use(express.json({limit: '32mb'}));
 app.use(express.static('static'));
 app.use(express.static('generated'));
 
+app.locals.appVersion = appVersion;
+
+// Get user agent
 app.use((req, res, next) => {
-  app.locals.appVersion = appVersion;
-  app.locals.browser = parseUA(req.get('User-Agent'), bcdBrowsers);
+  res.locals.browser = parseUA(req.get('User-Agent'), bcdBrowsers);
```
foolip, line 94: Isn't this used in some place that also needs updating?

PR #1582, opened 2021-10-16
Simplify release process instructions

queengooborg:
Now that we have the release script, the old instructions are no longer needed.  This PR replaces them.

(We could also keep them around as "legacy instructions" too?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ npm run clean
 
 ## Release process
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -182,14 +182,6 @@ npm run clean
 
 These are the manual steps to release and deploy a new version on https://mdn-bcd-collector.appspot.com/:
```
foolip, line 183: This can also be simplified to just "To release ..."

foolip (approved):

PR #1584, opened 2021-10-16
Use original formatting from custom tests; don't run Prettier in build.js

### LINKED ISSUE #936: Most of the build time is spent in `prettier.format()`
As the number of tests grow, `npm run build` is getting slower, and sometimes tests for build.js have been timing out after 5 seconds on macOS on GitHub Actions.

After a quick profile, it seems like this single line accounts for most of the time spent:
https://github.com/foolip/mdn-bcd-collector/blob/7fc154bf75512ee67fb90e6a8ffe79f7c7788ccf/build.js#L58

Skipping that results in a build time of ~1.2s on my machine (macOS), compared to ~4.7s without changes. So about 75% of the time is spend formatting.

This can probably be fixed pretty easily.

queengooborg:
This PR fixes #936 by keeping the original newlines within custom tests and not running Prettier.  We lose the ability to detect syntax errors or other malformed test issues beforehand (at least, not without writing a script to check it), but we do gain a major performance boost.


foolip:
> Nope, this one doesn't need any updating; it's all good to go! The custom test updates were just to the IDL, it looks like!

This seemed a bit too good to be true, so I double checked. There are remaining cases of `>-` that have been landed in the meantime. I'll fix them.

unittest/unit/build.js
```diff
@@ -145,26 +136,29 @@ describe('build', () => {
...
+          '(function() {\n  var a = 1;\n  if (!a) {\n    return false;\n  }\n  var instance = a;\n  return !!instance;\n})();'
         );
+      });
 
+      it('valid import: two imports', () => {
+        // XXX The "var b = a;" should be indented...
         assert.equal(
           getCustomTestAPI('import2'),
-          '(function () {\n  var a = 1;\n  if (!a) {\n    return false;\n  }\n  var b = a;\n  if (!b) {\n    return false;\n  }\n  var instance = b;\n  return !!instance;\n})();'
+          '(function() {\n  var a = 1;\n  if (!a) {\n    return false;\n  }\nvar b = a;\n  if (!b) {\n    return false;\n  }\n  var instance = b;\n  return !!instance;\n})();'
```
foolip, line 147: The indentation here seems to have regressed, this string represents this:
```js
(function() {
 var a = 1;
 if (!a) {
 return false;
 }
var b = a;
 if (!b) {
 return false;
 }
 var instance = b;
 return !!instance;
})();
```

The `return` statements and `var b = a` are indented too little.

foolip (changes requested):
All looks good apart from the one bad indentation I spotted. I also tried this locally and looked at some custom tests without finding anything else wrong.

unittest/unit/build.js
queengooborg: Yeah...I can't figure out what's going on with the indentations on imports.  I added an XXX comment for now to remember to look into it at a later time (and I'll open an issue too once this PR's merged).

foolip: I see, I didn't see the comment on line 144.

foolip (approved):

foolip:
@queengooborg there are conflicts to resolve again now :)

Issue #1591, opened 2021-10-19
Tests get stuck on "Running tests for ServiceWorker..." in Chrome 34


foolip:
Running http://mdn-bcd-collector.appspot.com/tests/api/TextMetrics?exposure=Window in Chrome 34 on BrowserStack, it gets stuck looking like this:
![image](https://user-images.githubusercontent.com/498917/137928133-3a351805-7771-49f0-b400-57ee5336829b.png)

In the console I see this:
![image](https://user-images.githubusercontent.com/498917/137928255-82445170-6be5-4f4f-895b-b41f174c7f85.png)

I'm not sure why there's an attempt to run service worker tests at all, since there aren't any for this page.

@queengooborg have you seen this?

queengooborg:
Hmm, I think the issue of the tests getting stuck might have been a fluke.  I've run this same test a couple of times in both BrowserStack and SauceLabs in Chrome 34 and couldn't get it to freeze.

Regarding the logs stating the tests are running in exposure sets that there aren't any tests, I'll send a PR to fix that up!

PR #1595, opened 2021-10-20
Only report "running tests for X" if X has tests to run

queengooborg:
This PR fixes the logging bug that was mentioned in #1591.

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -381,9 +381,8 @@
   }
 
   function runWindow(callback) {
-    setCurrentExposure('Window');
-
     if (pending.Window) {
+      setCurrentExposure('Window');
       runTests(pending.Window, callback);
     } else {
       callback([]);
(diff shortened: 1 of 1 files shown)


static/resources/harness.js
```diff
@@ -381,7 +383,7 @@
   }
 
   function runWindow(callback) {
-    setCurrentExposure('Window');
+    setCurrentExposure('Window', !!pending.Window);
```
foolip, line 386: Can we just not call `setCurrentExposure` if there aren't any pending tests, by moving this call into the if statement?

queengooborg: I tried that at first but it ended up causing other issues.  The code that catches if an exposure set is run twice depends on the exposure sets' run functions to be called in that order, and if we skip one, we end up with the tests just freezing (since the collector only lets the first completion of an exposure set continue on).

foolip: If the problem is with `navigator.mediaDevices.getUserMedia()` returning the same promise multiple times or if the `.then()` callback of certain promises gets called multiple times, is there a more local workaround for those odd situations that doesn't affect the overall harness?

queengooborg:
Closing in favor of #1703.

PR #1653, opened 2021-11-24
Add custom tests for Geolocation APIs

queengooborg:
This PR adds custom tests for the Geolocation APIs.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -795,6 +795,22 @@ api:
       var instance = reusableInstances.audioContext.createGain();
   GamepadEvent:
     __base: var instance = new GamepadEvent('gamepadconnected');
+  Geolocation:
+    __base: var instance = navigator.geolocation;
+  GeolocationCoordinates:
+    __base: >-
+      <%api.GeolocationPosition:posCallback%>
+      function posCallback(pos) {
+        callback(pos.coords);
+      }
+  GeolocationPosition:
+    __base: >-
+      <%api.Geolocation:geo%>
+      geo.getCurrentPosition(function(pos) {
+        callback(pos);
+      }, function(err) {
+        fail(err.message);
+      });
   HashChangeEvent:
     __base: >-
       var instance;


custom-tests.yaml
```diff
@@ -795,6 +795,22 @@ api:
...
+  GeolocationCoordinates:
+    __base: >-
+      <%api.GeolocationPosition:posCallback%>
+      function posCallback(pos) {
+        callback(pos.coords);
+      }
+  GeolocationPosition:
+    __base: >-
+      <%api.Geolocation:geo%>
+      geo.getCurrentPosition(function(pos) {
```
foolip, line 809: The reuse doesn't really help here I think, it's less code and easier to understand to inline:

```suggestion
      navigator.geolocation.getCurrentPosition(function(pos) {
```

custom-tests.yaml
```diff
@@ -795,6 +795,22 @@ api:
       var instance = reusableInstances.audioContext.createGain();
   GamepadEvent:
     __base: var instance = new GamepadEvent('gamepadconnected');
+  Geolocation:
+    __base: var instance = navigator.geolocation;
+  GeolocationCoordinates:
+    __base: >-
+      <%api.GeolocationPosition:posCallback%>
```
foolip, line 802: This ends up generating a working test, but it's weird to see `posCallback` as the variable name for what would be an instance in an non-callback tests. It looks like this is the first such case, or at least there are no cases of "callback%" in this file.

I have another suggestion (instance reuse) that might avoid the issue.

foolip (changes requested):
It looks like running these tests is pretty slow, it ends up calling `navigator.geolocation.getCurrentPosition()` many times. How about a reused instance of `GeolocationPosition` to avoid this?

custom-tests.yaml
queengooborg: It does simplify the code, but I wanted to also make sure that this test returns `false` if `navigator.geolocation` doesn't exist (if `Geolocation` isn't available, there's no way `GeolocationPosition` can be).

custom-tests.yaml
queengooborg: This would be the first test that imports a callback test, and with the way callback tests are implemented right now I couldn't figure out a better way than this for now.  I'd love to be able turn this into a simpler test in the future, but for the time being I feel this should be alright?

foolip: I suspect most of this was all supported from the start. How about having just one custom test so that there's just one call to the API, and doing manual testing if some of the data is uncertain?

I'm hesitant both because of the way callback reuse looks, and because of the many calls this will result in, adding to test execution time.

queengooborg: Alright, I was thinking about this one again, and I had some ideas:

---

# Custom test creation

Digging into the file system API has made me rethink imported callback tests somewhat, and I think I've got an idea for how to simplify them.  While it won't help with the performance as much, it'll help with code construction.

I'm thinking that it would make sense to automatically perform the function wrapping on the build end.  Basically, if a callback function is imported, take the remaining code and wrap it in a function we'll call `after_var()` (`var` being the variable name).  From the perfective of our custom tests file, variables returned through a callback will be no different than a regular variable.

In other words, this:
```js
<%api.GeolocationPosition:posCallback%>
function posCallback(pos) {
  callback(pos.coords);
}
```

...becomes this:
```js
<%api.GeolocationPosition:pos%>
callback(pos.coords);
```

...and the generated code remains roughly the same.

# Performance improvements

## In general

What would be nice to do is have the repeated code performed for each base feature and subfeature only run _once,_ store its return value, and reuse it in all of the subfeatures.  This is something that warrants a collector v5 release, and would be a pretty big change.

## For this test

I ran some metrics to determine just how much of a performance impact we're getting by adding these tests.  I ran the tests three times and used the console logs to time test start to test completion (pre-server postings), and got the following results (using Chrome 96 on macOS):

```
*Without Geolocation Tests*
1st: 14.359 - 08.712 = 5.647
2nd: 33.762 - 28.558 = 5.204
3rd: 11.355 - 6.036  = 5.319
Avg: (5.647 + 5.204 + 5.319) / 3 = 5.390

*With Only GeolocationPosition*
1st: 38.047 - 32.980 = 5.067
2nd: 23.445 - 18.216 = 5.229
3rd: 59.036 - 53.962 = 5.074
Avg: (5.067 + 5.229 + 5.074) / 3 = 5.123

*Without*
1st: 26.842 - 22.932 = 3.910
2nd: 27.705 - 24.182 = 3.523
3rd: 54.779 - 51.001 = 3.778
Avg: (3.910 + 3.523 + 3.778) / 3 = 3.737
```

So we're not adding an insignificant amount of latency with these tests, but I feel that it might be an acceptable amount...?

foolip: Which two numbers should I compare there? What's the difference between Without Geolocation Tests and Without?

queengooborg: Whoops, didn't mean for the top one to say "Without" -- fixed!

PR #1656, opened 2021-11-25
Add test for CanvasRenderingContext2D.drawFocusIfNeeded.path_parameter

queengooborg:


custom-tests.yaml
```diff
@@ -298,6 +298,11 @@ api:
       };
       var instance = canvas.getContext('2d');
     __additional:
+      drawFocusIfNeeded.path_parameter: >-
+        var path = new Path2D();
+        var el = document.createElement('p');
+        instance.drawFocusIfNeeded(path, el);
```
foolip, line 304: If the detection is based on this throwing an exception, can you do an explicit try-catch?

foolip (changes requested):

PR #1660, opened 2021-11-26
Add custom test for RTCIceCandidate API

queengooborg:


custom-tests.yaml
```diff
@@ -1550,6 +1550,8 @@ api:
       var instance = new RTCDataChannelEvent('datachannel', {channel: channel});
   RTCDTMFToneChangeEvent:
     __base: "var instance = new RTCDTMFToneChangeEvent('tonechange', {tone: ''});"
+  RTCIceCandidate:
+    __base: var instance = new RTCIceCandidate();
```
foolip, line 1554: Before https://source.chromium.org/chromium/chromium/src/+/d1d6586b3beca9bfdf3f2e1b9951e4bbf074604c the argument was required.

https://source.chromium.org/chromium/chromium/src/+/8836d50a6ae1da37f8f514d59a0ab144bf887cb5 is the commit that introduced `RTCIceCandidate` and shows what the original signature was. I'm not sure which parameters were required.

foolip (changes requested):

custom-tests.yaml
queengooborg: Oddly enough, trying to construct this in the latest Chrome without the parameters I just added seems to fail, so they're still required in Chrome apparently...huh.

Issue #1668, opened 2021-12-02
Improving imported callback tests


queengooborg:
Digging into the file system API has made me rethink imported callback tests somewhat, and I think I've got an idea for how to simplify them.  While it won't help with the performance as much, it'll help with code construction.

I'm thinking that it would make sense to automatically perform the function wrapping on the build end.  Basically, if a callback function is imported, take the remaining code and wrap it in a function we'll call `after_var()` (`var` being the variable name).  From the perfective of our custom tests file, variables returned through a callback will be no different than a regular variable.

In other words, this:
```js
<%api.GeolocationPosition:posCallback%>
function posCallback(pos) {
  callback(pos.coords);
}
```

...becomes this:
```js
<%api.GeolocationPosition:pos%>
callback(pos.coords);
```

...and the generated code remains roughly the same.

_Originally posted by @queengooborg in https://github.com/foolip/mdn-bcd-collector/pull/1653#discussion_r761015758_

PR #1673, opened 2021-12-06
Add custom tests for file system APIs

### LINKED ISSUE #1670: Create custom tests for File System API
https://github.com/mdn/browser-compat-data/pull/13514 and https://github.com/mdn/browser-compat-data/pull/13515 showed me that we need to write some custom tests for the file system API: https://developer.mozilla.org/en-US/docs/Web/API/FileSystem

Unfortunately at first glance, it seems this involves lots, and LOTS, of callbacks.  Maybe get #1668 finished first...?

queengooborg:
This PR adds custom tests for the file system APIs, which fixes #1670.  As a part of these additions, a slight adjustment to `build.js` has been made to make importing a callback as the instance a little easier.

Note: the custom test code looks a little ugly right now, but that will be resolved by #1668 later down the line.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -770,8 +770,8 @@ api:
     __base: >-
       window.webkitRequestFileSystem(TEMPORARY, 8*1024 /* 8KB */, function(fs) {
         callback(fs);
-      }, function(event) {
-        fail(event.message);
+      }, function(error) {
+        fail(error.message);
       });
   FileSystemEntry:
     __base: <%api.FileSystemDirectoryEntry:callback%>
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -766,6 +766,37 @@ api:
       var instance = new FetchEvent('fetch', { request: req });
   FileReader:
     __base: var instance = new FileReader();
+  FileSystem:
+    __base: >-
+      window.webkitRequestFileSystem(TEMPORARY, 8*1024 /* 8KB */, function(fs) {
+        callback(fs);
+      }, function(event) {
+        fail(event.message);
+      });
+  FileSystemEntry:
+    __base: <%api.FileSystemDirectoryEntry:callback%>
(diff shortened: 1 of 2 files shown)


custom-tests.yaml
```diff
@@ -766,6 +766,37 @@ api:
       var instance = new FetchEvent('fetch', { request: req });
   FileReader:
     __base: var instance = new FileReader();
+  FileSystem:
+    __base: >-
+      window.webkitRequestFileSystem(TEMPORARY, 8*1024 /* 8KB */, function(fs) {
+        callback(fs);
+      }, function(event) {
+        fail(event.message);
```
foolip, line 774:
```suggestion
      }, function(error) {
        fail(error.message);
```

custom-tests.yaml
```diff
@@ -766,6 +766,37 @@ api:
       var instance = new FetchEvent('fetch', { request: req });
   FileReader:
     __base: var instance = new FileReader();
+  FileSystem:
+    __base: >-
+      window.webkitRequestFileSystem(TEMPORARY, 8*1024 /* 8KB */, function(fs) {
```
foolip, line 771: Gecko also has an unprefixed `FileSystem` interface which you can get from `FileSystemEntry` objects. Can you test if there's a way to get to a `FileSystem` instance in Firefox?

custom-tests.yaml
```diff
@@ -766,6 +766,37 @@ api:
...
+        callback(fs);
+      }, function(event) {
+        fail(event.message);
+      });
+  FileSystemEntry:
+    __base: <%api.FileSystemDirectoryEntry:callback%>
```
foolip, line 777: Can we get a `FileSystemEntry` instance from `DataTransferItem.prototype.webkitGetAsEntry()` or `HTMLInputElement.prototype.webkitEntries`?

PR #1676, opened 2021-12-06
Format code on page using highlight.js

queengooborg:
This PR adds code syntax highlighting to the displayed codeblocks on the page.  Note: this does add another script to the page, but it's a small one (and I plan to set up a minified, JavaScript-only version in the future to reduce the script size even more).


views/tests.ejs
```diff
@@ -42,6 +42,11 @@
 <!-- TODO: https://github.com/foolip/mdn-bcd-collector/issues/407 -->
 <script src="/resources/json3.min.js"></script>
 <script src="/resources/harness.js"></script>
+
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/11.3.1/styles/stackoverflow-dark.min.css">
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/11.3.1/styles/stackoverflow-light.min.css" media="(prefers-color-scheme: light)">
+<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/11.3.1/highlight.min.js"></script>
```
foolip, line 48: https://github.com/foolip/mdn-bcd-collector/issues/407 (see json3.min.js above) applies here too. Can you arrange for this script to be loaded only after tests have finished running?

foolip (changes requested):

views/tests.ejs
queengooborg: Sure thing!  Do you want me to make the changes for highlight.js here, or tackle both scripts at once?

static/resources/harness.js
```diff
@@ -693,6 +693,43 @@
     }
   }
 
+  function loadHighlightJs(callback) {
+    try {
+      // Load dark (main) style
+      var darkStyle = document.createElement('link');
+      darkStyle.rel = 'stylesheet';
+      darkStyle.href =
+        '//cdnjs.cloudflare.com/ajax/libs/highlight.js/11.3.1/styles/stackoverflow-dark.min.css';
```
foolip, line 708: Can we put everything in resources/ to avoid this?

queengooborg: I wanted to do that, but this turns out this was a bigger hurdle than I originally thought.  We'll have to build our own bundle, which while I'd love to do that (since we can pick and choose syntax languages to minify our bundle to our perfect needs), I figured it would be better to use the pre-built bundle for now!

foolip (approved):
🎄

PR #1681, opened 2021-12-09
Handle Android WebView 4.4.3 in ua-parser

queengooborg:
Looks like because we're stripping the patch version from the version number, we're accidentally mapping 4.4.3 to 4.4 -- see https://github.com/foolip/mdn-bcd-results/pull/2719.﻿


foolip (changes requested):
Looks good, but can you add a test for this case?

foolip (approved):

PR #1690, opened 2021-12-15
Load JSON3 polyfill only when needed

queengooborg:
This PR updates our code to only load JSON3 when it's needed, in two ways:

- Only load JSON3 _after_ all the tests; it's not needed for worker tests at all since workers were added after JSON support
- Check if we need to load the polyfill in the first place; if the browser has the JSON support we need, don't load the script

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -764,50 +764,70 @@
     resultsEl.appendChild(resultEl);
   }
 
+  function sendReport(results) {
+    var body = JSON.stringify(results);
+
+    var client;
+    if ('XMLHttpRequest' in self) {
+      client = new XMLHttpRequest();
+    } else if ('ActiveXObject' in self) {
+      client = new ActiveXObject('Microsoft.XMLHTTP');
+    }
+
+    if (!client) {
+      updateStatus(
+        'Cannot upload results: XMLHttpRequest is not supported.',
+        'error-notice'
+      );
+      return;
+    }
+
+    var resultsURL =
+      (location.origin || location.protocol + '//' + location.host) +
+      '/api/results?for=' +
+      encodeURIComponent(location.href);
+
+    client.open('POST', resultsURL);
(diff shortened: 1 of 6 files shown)


static/resources/harness.js
```diff
@@ -764,50 +764,70 @@
...
+
+    try {
+      if ('JSON' in self && 'parse' in JSON) {
+        sendReport(results);
+      } else {
+        // Load JSON polyfill if needed
+        var polyfill = document.createElement('script');
+        polyfill.src = '/resources/json3.min.js';
+
+        polyfill.addEventListener(
```
foolip, line 821: Does this work as far back as the previous code? `addEventListener` was added in IE9, right?

foolip (commented):
Yay, I like this! A question about ancient browsers though...

static/resources/harness.js
queengooborg: I've double-checked to make sure this doesn't cause an issue in older browsers  -- it functions in IE8, though it seems IE6 and IE7 are having a little issue.  Looking into it now!

queengooborg: Alright, looks like IE doesn't support the `load` event before IE9, so I've added a workaround where the report will be sent after a small delay instead.  Fortunately this is an IE-only issue, other browsers aren't affected!

static/resources/harness.js
```diff
@@ -764,50 +764,73 @@
...
-        client = new XMLHttpRequest();
-      } else if ('ActiveXObject' in self) {
-        client = new ActiveXObject('Microsoft.XMLHTTP');
-      }
+        if ('onload' in polyfill) {
+          polyfill.onload = function () {
+            sendReport(results);
+          };
+        } else {
+          // If we can't determine when the polyfill loads, use a delay
```
foolip, line 826: If this turns out to be unreliable (due to import taking we could have the imported script call a function we've defined. But only if it becomes a problem in practice.

foolip (approved):

PR #1703, opened 2021-12-21
Simplify code to check for double-run of test exposures

queengooborg:
This PR simplifies the code we use to check if an exposure was run twice, by completely removing it.  Instead, we ensure that the callback can only be called once.﻿


foolip (approved):
So much red 🤩

PR #1709, opened 2021-12-23
Release script: simplify test list

queengooborg:
This PR updates the changed test lists by consolidating the lists so that all the children won't be listed if the parent has been modified (i.e. if 'api.AbortController' was added, 'api.AbortController.someProperty' won't be included in the list)

### DIFF
--- a/release.js
+++ b/release.js
@@ -164,13 +164,12 @@ const getTestChanges = () => {
         changed = changed.filter(simplifyTestChangesList);
 
         ctx.testChanges =
-          '\n#### Added\n<details>\n' +
-          added.join('<br />\n') +
-          '\n</details>\n\n#### Removed\n<details>\n' +
-          removed.join('<br />\n') +
-          '\n</details>\n\n#### Changed\n<details>\n' +
-          changed.join('<br />\n') +
-          '\n</details>\n';
+          '\n#### Added\n\n' +
+          added.map((x) => '- ' + x).join('\n') +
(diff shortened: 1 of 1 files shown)


release.js
```diff
@@ -135,22 +149,27 @@ const getTestChanges = () => {
         const oldTestKeys = Object.keys(oldTests);
         const newTestKeys = Object.keys(newTests);
 
-        const added = newTestKeys.filter((k) => !oldTestKeys.includes(k));
-        const removed = oldTestKeys.filter((k) => !newTestKeys.includes(k));
-        const changed = [];
+        const added = newTestKeys
+          .filter((k) => !oldTestKeys.includes(k))
+          .filter(simplifyTestChangesList);
```
foolip, line 154: This will be O(n^2) because `list.includes` is called for east list member. It doesn't matter because the lists will be short, but doing it fast isn't really harder. I'd suggest a helper function that takes a list and returns a new list. First create a set from the list, then for each member of the set check if a shorter version of it is in the set and remove the longer ones. Then convert the set to a list and return.

foolip (changes requested):
I don't see anything about `<details>` in the diff?

queengooborg:
Sorry, forgot to update the PR -- I removed the `<details>` elements and opted for a Markdown list instead, because if the list of changed tests really gets that long, that's an indication of a problem.  (Well, aside from the formatting changes of v5.0.)

PR #1716, opened 2021-12-27
Generate toString tests for named stringifiers in IDL

queengooborg:
This PR updates the build script to add `toString` tests for any _named_ attributes with the `stringifier` special attribute.  This case can be found in the `HTMLHyperlinkElementUtils` mixin and `WorkerLocation`﻿interface.

Fixes #1057.

PR #1717, opened 2021-12-27
Add custom tests for HTML[Anchor/Area]Element/WorkerLocation toString

queengooborg:
This PR adds custom tests for the `toString` method of the `HTMLAnchorElement`, `HTMLAreaElement`, and `WorkerLocation` interfaces.  This is a sister PR to #1716, which updates `build.js` to include these members.﻿

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -827,9 +827,15 @@ api:
   HTMLAnchorElement:
     __base: var instance = document.createElement('a');
     __test: return bcd.testObjectName(instance, 'HTMLAnchorElement');
+    toString: |-
+      instance.href = 'https://mdn-bcd-collector.appspot.com/';
+      return instance.toString() == instance.href;
   HTMLAreaElement:
     __base: var instance = document.createElement('area');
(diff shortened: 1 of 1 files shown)


custom-tests.yaml
```diff
@@ -2437,6 +2443,9 @@ api:
     __base: var instance = self;
   WorkerLocation:
     __base: var instance = location;
+    toString: |-
+      instance.href = 'https://mdn-bcd-collector.appspot.com/';
```
foolip, line 2447: Does this really do anything? This is read only in workers so setting it should do nothing.

queengooborg: Good catch, fixed!

Earthman100: Its for 4d and lint.yml used for _CI.yml to Blank.yml for Create Main.yml on create from 4d to 3d as 2d -8bit process runner back to 4d gui

Earthman100: I have a lot of repos that I have forked to prevent changing source as to the 4d gui and the differences between 4d and 3d computers, and of course the long debated debutes of 4d and 3d related to 2d objects in a 3d computer as inflated and how thats not 4d

PR #1756, opened 2022-01-25
Add support for wildcard exposures

queengooborg:
This PR adds support for [wildcard exposure sets](https://webidl.spec.whatwg.org/#ref-for-dfn-xattr-wildcard%E2%91%A0).  While this feature is rarely used (currently only in a WICG draft spec, https://wicg.github.io/compression/#compression-stream), it was included in Webref/idl v3, which is blocking #1747.  This maps the wildcard to the four main exposure sets (window, worker, shared worker, and service worker).

Additionally, this renames the "globals" variable within `getExposureSet()` to `exposure` for better naming.

### DIFF
--- a/build.js
+++ b/build.js
@@ -429,9 +429,9 @@ const getExtAttr = (node, name) => {
   return node.extAttrs && node.extAttrs.find((i) => i.name === name);
 };
 
-// https://webidl.spec.whatwg.org/#dfn-exposure-set
+// https://webidl.spec.whatwg.org/#Exposed
 const getExposureSet = (node) => {
-  // step 6-8
+  // step 6-8 of https://webidl.spec.whatwg.org/#dfn-exposure-set
   const attr = getExtAttr(node, 'Exposed');
   if (!attr) {
     throw new Error(
@@ -445,6 +445,16 @@ const getExposureSet = (node) => {
       break;
     case 'identifier-list':
       for (const {value} of attr.rhs.value) {
+        exposure.add(value);
+      }
+      break;
+    case '*':
+      for (const value of [
+        'Window',
+        'Worker',
+        'SharedWorker',
+        'ServiceWorker'
+      ]) {
         exposure.add(value);
       }
       break;
--- a/build.js
+++ b/build.js
@@ -438,14 +438,14 @@ const getExposureSet = (node) => {
       `Exposed extended attribute not found on ${node.type} ${node.name}`
     );
   }
-  const globals = new Set();
+  const exposure = new Set();
   switch (attr.rhs.type) {
     case 'identifier':
-      globals.add(attr.rhs.value);
+      exposure.add(attr.rhs.value);
       break;
     case 'identifier-list':
       for (const {value} of attr.rhs.value) {
-        globals.add(value);
+        exposure.add(value);
       }
       break;
     /* istanbul ignore next */
@@ -453,12 +453,12 @@ const getExposureSet = (node) => {
       throw new Error(`Unexpected RHS for Exposed extended attribute`);
   }
 
-  if (globals.has('DedicatedWorker')) {
-    globals.delete('DedicatedWorker');
-    globals.add('Worker');
+  if (exposure.has('DedicatedWorker')) {
+    exposure.delete('DedicatedWorker');
+    exposure.add('Worker');
   }
 
-  return globals;
+  return exposure;
 };
 
 const validateIDL = (ast) => {
--- a/build.js
+++ b/build.js
@@ -348,6 +348,7 @@ const flattenIDL = (specIDLs, customIDLs) => {
   const scopes = ast
     .filter((dfn) => getExtAttr(dfn, 'Global'))
     .map((dfn) => dfn.name.replace('GlobalScope', ''));
+  scopes.push('Worker', 'Worklet');
 
   return {ast, globals, scopes};
 };
--- a/build.js
+++ b/build.js
@@ -473,7 +473,6 @@ const getExposureSet = (node, scopes) => {
 
   for (const e of exposure) {
     if (!scopes.includes(e)) {
-      console.log(scopes);
       throw new Error(
         `${node.type} ${node.name} is exposed on ${e} but ${e} is not a valid scope`
       );
--- a/build.js
+++ b/build.js
@@ -345,12 +345,34 @@ const flattenIDL = (specIDLs, customIDLs) => {
   );
 
   // Get all possible scopes
-  const scopes = ast
-    .filter((dfn) => dfn.name && getExtAttr(dfn, 'Global'))
-    .map((dfn) =>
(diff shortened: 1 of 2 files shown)


build.js
```diff
@@ -429,36 +429,46 @@ const getExtAttr = (node, name) => {
...
       break;
     case 'identifier-list':
       for (const {value} of attr.rhs.value) {
-        globals.add(value);
+        exposure.add(value);
+      }
+      break;
+    case '*':
+      for (const value of [
+        'Window',
```
foolip, line 453: Per spec this should be the set of all interfaces which have the `[Global]` extended attribute. That includes https://webaudio.github.io/web-audio-api/#audioworkletglobalscope and other things not on this list. We don't generate any tests for worklets yet, but I think we should still collect all of the valid global scopes from parsing all of the IDL. We'd use that set here, but we can also use it to validate that there aren't any typos in global scope names anywhere.

foolip (changes requested):

build.js
queengooborg: Just got that logic added in to grab everything with a `[Global]` extended attribute, with a bit of remapping, and use that as the value for wildcard exposures.  I've also added in the validation logic for an extra touch!

build.js
```diff
@@ -344,7 +344,15 @@ const flattenIDL = (specIDLs, customIDLs) => {
     (dfn) => dfn.type !== 'includes' && dfn.type !== 'interface mixin'
   );
 
-  return {ast, globals};
+  // Get all possible scopes
```
foolip, line 347: This part isn't quite right. We don't need to strip 'GlobalScope' if we use the identifiers in the Global extended attribute. Here's what I got after some tinkering:
```js
  const scopes = new Set();
  for (const dfn of ast) {
    const attr = getExtAttr(dfn, 'Global');
    if (!attr) {
      continue;
    }
    // Special case RTCIdentityProviderGlobalScope since it doesn't use the
    // Global extended attribute correctly:
    // https://github.com/w3c/webrtc-identity/pull/36
    if (dfn.name === 'RTCIdentityProviderGlobalScope') {
      scopes.add('RTCIdentityProvider');
      continue;
    }
    switch (attr.rhs.type) {
      case 'identifier':
        scopes.add(attr.rhs.value);
        break;
      case 'identifier-list':
        for (const {value} of attr.rhs.value) {
          scopes.add(value);
        }
        break;
      default:
        throw new Error(`Unable to parse Global extended attribute of ${dfn.name}`);
    }
  }
```

Can you add that and adapt the tests?

build.js
```diff
@@ -429,36 +437,49 @@ const getExtAttr = (node, name) => {
...
     throw new Error(
       `Exposed extended attribute not found on ${node.type} ${node.name}`
     );
   }
-  const globals = new Set();
+  const exposure = new Set();
   switch (attr.rhs.type) {
     case 'identifier':
-      globals.add(attr.rhs.value);
+      exposure.add(attr.rhs.value.replace('GlobalScope', ''));
```
foolip, line 452: Is `.replace('GlobalScope', '')` needed here? We should be processing an extended attribute like `[Exposed=Worker]` here, and the 'GlobalScope' suffix isn't part of the identifiers used in the extended attribute.

build.js
```diff
@@ -429,36 +437,49 @@ const getExtAttr = (node, name) => {
...
       break;
     /* istanbul ignore next */
     default:
       throw new Error(`Unexpected RHS for Exposed extended attribute`);
   }
 
-  if (globals.has('DedicatedWorker')) {
-    globals.delete('DedicatedWorker');
-    globals.add('Worker');
+  if (exposure.has('DedicatedWorker')) {
```
foolip, line 469: This is a workaround for something. Can you link to https://github.com/foolip/mdn-bcd-collector/pull/811 and explain why this is needed? At first glance it just seems wrong to do this replacement.

unittest/unit/build.js
```diff
@@ -671,20 +696,63 @@ describe('build', () => {
...
       const interfaces = ast.filter((dfn) => dfn.type === 'interface');
-      const exposureSet = getExposureSet(interfaces[0]);
+      const exposureSet = getExposureSet(interfaces[0], scopes);
       assert.hasAllKeys(exposureSet, ['Worker']);
     });
+
+    it('"GlobalScope" stripped from exposure set names', () => {
+      const specIDLs = {
+        first: WebIDL2.parse(
+          `[Exposed=AudioWorkletGlobalScope]
```
foolip, line 706: `[Exposed=AudioWorklet]` is the syntax that's actually used in specs, so I think this test can probably be removed together with the code that does the stripping.

foolip (changes requested):

build.js
queengooborg: This is to handle the same issue for `RTCIdentityProviderGlobalScope`.  Since this is an issue with the spec and we're special casing it during our search for `Globals`, we can special-case it here too!

build.js
```diff
@@ -344,7 +344,24 @@ const flattenIDL = (specIDLs, customIDLs) => {
...
+    // https://github.com/w3c/webrtc-identity/pull/36
+    if (dfn.name === 'RTCIdentityProviderGlobalScope') {
+      scopes.add('RTCIdentityProvider');
+      continue;
+    }
+
+    const attr = getExtAttrSet(dfn, 'Global', false);
+    for (const s of attr) {
+      scopes.add(s);
+    }
```
foolip, line 361: To avoid the `throwError` argument, we can make this code just a little bit more complicated:

```suggestion
    const attr = getExtAttrSet(dfn, 'Global');
    if (attr) {
      for (const s of attr) {
        scopes.add(s);
      }
    }
```

It should be a simplification overall.

build.js
```diff
@@ -429,36 +446,80 @@ const getExtAttr = (node, name) => {
   return node.extAttrs && node.extAttrs.find((i) => i.name === name);
 };
 
-// https://webidl.spec.whatwg.org/#dfn-exposure-set
-const getExposureSet = (node) => {
-  // step 6-8
-  const attr = getExtAttr(node, 'Exposed');
+const getExtAttrSet = (node, name, throwError = true) => {
```
foolip, line 449:
```suggestion
const getExtAttrSet = (node, name) => {
```

build.js
```diff
@@ -429,36 +446,80 @@ const getExtAttr = (node, name) => {
...
-    throw new Error(
-      `Exposed extended attribute not found on ${node.type} ${node.name}`
-    );
+    if (throwError) {
+      throw new Error(
+        `${name} extended attribute not found on ${node.type} ${node.name}`
+      );
+    } else {
+      return new Set();
+    }
```
foolip, line 458:
```suggestion
    return null;
```

foolip (changes requested):

foolip (approved):

PR #1777, opened 2022-02-03
Switch to Yarn as package manager

### LINKED ISSUE #1786: Switch from NPM to Yarn
The majority of MDN Web Docs uses Yarn as their package manager, so I'm thinking we should as well.  Doing so will also allows us to simultaneously support development in NodeJS v14 and NodeJS v15+ without worrying about the version for `package-lock.json`.

queengooborg:
The majority of MDN Web Docs uses Yarn as their package manager, so I'm thinking we should as well.  Doing so will also allows us to simultaneously support development in NodeJS v14 and NodeJS v15+ without worrying about the version for `package-lock.json`.  Fixes #1786.


queengooborg:
Note: the failing build is due to an issue with `nodegit` not being compatible in NodeJS v15 or higher.  I'm going to fix this in a separate PR!

foolip:
@ddbeck are you happy with the use of Yarn in other MDN projects? I've never used it, but I do wonder if it has a bright future given that you have to go through extra steps to install it, compared to just using npm. I'm happy to go with whatever you'd like, but want to make sure you actually like it and want to see more of it :)

queengooborg:
I opened #1786 as we should come to a conclusion before a PR is opened.  Keeping this open will result in continuous merge conflicts as dependencies are updated.

queengooborg:
Closing this because it was decided that NPM would be used for all of MDN.

PR #1784, opened 2022-02-04
Update unittests

queengooborg:
This PR updates the unittests to add more coverage, as well as properly ignore blocks of code that are too challenging/impossible/not worthwhile to test.


foolip:
@queengooborg sorry I neglected this. Can you rebase?

queengooborg:
Actually, can you help with that?  I can't modify v1 package locks now due to a [bug with NPM v6](https://stackoverflow.com/questions/69452504/zlib-error-when-attempting-to-run-npm-install-or-yarn)...

unittest/app/app.js
```diff
@@ -249,6 +249,11 @@ describe('/tests/', () => {
     assert.equal(res.status, 200);
   });
 
+  it('get all tests, ignore CSS', async () => {
+    const res = await agent.get('/tests/?ignore=css');
+    assert.equal(res.status, 200);
```
foolip, line 254: Should something be asserted here that's different from the previous step? As it is, this test would pass even if the `ignore` parameter is unused.

unittest/unit/build.js
```diff
@@ -14,7 +14,7 @@
 
 'use strict';
 
-import chai, {assert, expect} from 'chai';
+import chai, {assert} from 'chai';
```
foolip, line 17: Yay, I like `assert` better than `expect` :)

unittest/unit/build.js
```diff
@@ -1302,28 +1357,39 @@ describe('build', () => {
...
+          properties: {
+            grid: {}
+          }
         }
-      }
-    };
+      };
 
-    const customCSS = {
+      const customCSS = {
```
foolip, line 1392: Accidental indentation here? Doesn't this fail some lint rule?

foolip (approved):

unittest/app/app.js
queengooborg: We should probably also check to make sure we're loading the correct tests as well in the previous two steps.  I'm not sure how to do it at this time though, so I've added some TODO comments to remind us later.

PR #1790, opened 2022-02-07
Fix Puppeteer codecov reporting

### LINKED ISSUE #1789: c8 doesn't work with puppeteer-to-istanbul
Looks like there's something going on between `c8` and `puppeteer-to-istanbul`, where it's not including test results in codecov reports.

queengooborg:
This PR fixes the codecov reporting coming from Puppeteer, so that code coverage is included in `c8`/`istanbul`.  Fixes #1789.


foolip (approved):
I see some harness.js coverage in the test output, so this seems to work!

PR #1805, opened 2022-02-14
Add the ability to test CSS property values to custom-css.json

### LINKED ISSUE #705: Check support for each CSS property keyword value
Currently the code only checks existence of properties but not the property values. We can probably extract the supported keywords from webref and use `CSS.supports` to check it.

gsnedders:
This also adds a lot of the more easily found values that either differ between {Chrome, Firefox, Safari} _or_ where entries already exist in BCD for them.

This almost certainly isn't quite what we want, but it's a starting point. Notably this produces significant changes to BCD data, as (unsurprisingly) plenty of this data is outdated.

Fixes #705, based on @foolip's earlier experience with #1221.

### DIFF
--- a/custom-css.json
+++ b/custom-css.json
@@ -9,21 +9,31 @@
     "-epub-text-emphasis": {},
     "-epub-text-emphasis-color": {},
     "-epub-text-emphasis-style": {},
-    "-epub-text-orientation": {},
-    "-epub-text-transform": {},
-    "-epub-word-break": {},
+    "-epub-text-orientation": {
+      "__values": ["sideways"]
+    },
+    "-epub-text-transform": {
+      "__values": ["capitalize", "full-size-kana", "full-width"]
+    },
+    "-epub-word-break": {
+      "__values": ["break-word", "keep-all"]
+    },
     "-epub-writing-mode": {},
     "-internal-text-autosizing-status": {},
     "-moz-animation": {},
     "-moz-animation-delay": {},
-    "-moz-animation-direction": {},
+    "-moz-animation-direction": {
+      "__values": ["alternate-reverse", "reverse"]
+    },
     "-moz-animation-duration": {},
     "-moz-animation-fill-mode": {},
     "-moz-animation-iteration-count": {},
     "-moz-animation-name": {},
     "-moz-animation-play-state": {},
     "-moz-animation-timing-function": {},
-    "-moz-appearance": {},
+    "-moz-appearance": {
+      "__values": ["auto", "menulist-button", "none", "textfield"]
+    },
     "-moz-backface-visibility": {},
     "-moz-background-inline-policy": {},
     "-moz-binding": {},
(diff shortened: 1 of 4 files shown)


queengooborg (changes requested):
This is a great PR, I'm enjoying this functionality and I think this is a great way to incorporate it!  I just have a few comments:

First, for line 838, we no longer have `propertySet` as a variable, so we'll need to do the following instead:
```js
for (const name of Array.from(properties.keys()).sort()) {
```

Second, I think that this test code should be thrown into `harness.js` and called via `bcd.testCSSProperty()`.  Since we're injecting this code so many times for each variation, the codeblock gets pretty big (about a half a page long).  I feel it would be helpful to put this into a reusable function!

gsnedders:
> This is a great PR, I'm enjoying this functionality and I think this is a great way to incorporate it! I just have a few comments:
> 
> First, for line 838, we no longer have `propertySet` as a variable, so we'll need to do the following instead:
> 
> ```js
> for (const name of Array.from(properties.keys()).sort()) {
> ```

…did I not commit before pushing? 🤦

> Second, I think that this test code should be thrown into `harness.js` and called via `bcd.testCSSProperty()`. Since we're injecting this code so many times for each variation, the codeblock gets pretty big (about a half a page long). I feel it would be helpful to put this into a reusable function!

Ah, this is the sorta feedback I was looking for! 

Looking at this again (sorry, I wrote it a while ago!), how do you feel about prefixes or what we should do with prefixes? Being able to go from something marked as supported with a prefix to being able to fill in its sub-values would be valuable, but I don't know it really makes sense to generate them for everywhere. One option would be to just generate tests for `-x-foo` if `-x-foo` exists in `custom-css.json` (i.e., append all the values into it).

queengooborg:
> Looking at this again (sorry, I wrote it a while ago!), how do you feel about prefixes or what we should do with prefixes?

Prefixes was something that we'd been looking into.  Currently, with APIs, our method is to add custom IDL with the prefixes, essentially treating prefixed interfaces/members as separate entities, to then resolve later in update-bcd.  Perhaps we could/should do something similar here?  @foolip, what do you think?

foolip:
For prefixed properties, I think we should generate separate tests and let the updater connect the dots. For prefixed property values, I guess we would add those as custom tests?

gsnedders:
> For prefixed properties, I think we should generate separate tests and let the updater connect the dots. For prefixed property values, I guess we would add those as custom tests?

but add them all explicitly, rather than implicitly?

gsnedders:
> Second, I think that this test code should be thrown into `harness.js` and called via `bcd.testCSSProperty()`. Since we're injecting this code so many times for each variation, the codeblock gets pretty big (about a half a page long). I feel it would be helpful to put this into a reusable function!

Done this now!

> > For prefixed properties, I think we should generate separate tests and let the updater connect the dots. For prefixed property values, I guess we would add those as custom tests?
> 
> but add them all explicitly, rather than implicitly?

And done this too!

queengooborg (approved):
Thank you for your hard work on this, this is LGTM!  👍

custom-css.json
```diff
@@ -312,49 +522,723 @@
...
     "color-profile": {},
     "color-rendering": {},
+    "color-scheme": {
+      "__values": ["only", "only_dark", "only_light"]
+    },
+    "column-fill": {
+      "__values": ["balance-all"]
+    },
     "column-progression": {},
-    "custom-property": {},
```
foolip, line 319: This removed a custom test by accident, restored in https://github.com/foolip/mdn-bcd-collector/pull/1945.

PR #1810, opened 2022-02-15
Don't render results automatically when there are more than 250

### LINKED ISSUE #1719: Don't render results automatically when there are lots
I've been using the `?selenium=true` GET variable on older browsers to prevent them from rendering all of the results when running all of the tests.  Some older devices, real and emulated, simply crash if it has to render too much, others freeze up.  At first, I was thinking of just bringing back the "Hide Results" checkbox, but then I realized that means our URL will have to change, which may cause some issues.

I propose we prevent the HTML rendering of the results if there are more than, say, 100, and instead render a button and text that says something along the lines of "There are X results, do you want to render them?"

queengooborg:
This PR fixes #1719.﻿

### DIFF
--- a/static/resources/harness.js
+++ b/static/resources/harness.js
@@ -858,12 +858,36 @@
 
     var resultsEl = document.getElementById('results');
 
-    if (resultsEl && !hideResults) {
+    function doRenderResults() {
       loadHighlightJs(function () {
         for (var i = 0; i < results.length; i++) {
           renderReportEl(results[i], resultsEl);
         }
       });
+    }
+
+    if (resultsEl && !hideResults) {
+      if (results.length > 250) {
+        var renderWarning = document.createElement('p');
+        renderWarning.innerHTML =
+          'There are ' +
+          results.length +
+          ' test results.<br />Displaying all results may cause your browser to freeze, especially on older browsers.<br />Display results anyways?';
+        resultsEl.appendChild(renderWarning);
+
+        var renderButton = document.createElement('button');
+        renderButton.innerText = 'Show Results';
+        resultsEl.appendChild(renderButton);
+
+        renderButton.onclick = function () {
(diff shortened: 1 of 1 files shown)


static/resources/harness.js
```diff
@@ -858,13 +858,37 @@
...
       });
     }
+
+    if (resultsEl && !hideResults) {
+      if (results.length > 250) {
+        var renderWarning = document.createElement('p');
+        renderWarning.innerHTML =
+          'There are ' +
+          results.length +
+          ' test results.<br />Displaying all results may cause your browser to freeze, especially on older browsers.<br />Display results anyways?';
```
foolip, line 875:
```suggestion
          ' test results.<br>Displaying all results may cause your browser to freeze, especially on older browsers.<br>Display results anyways?';
```

static/resources/harness.js
```diff
@@ -858,13 +858,37 @@
...
+        resultsEl.appendChild(renderWarning);
+
+        var renderButton = document.createElement('button');
+        renderButton.innerText = 'Show Results';
```
foolip, line 879:
```suggestion
        renderButton.innerHTML = 'Show Results';
```

This should allow the code to work on older Firefox, assuming BCD is correct about support.

static/resources/harness.js
```diff
@@ -858,13 +858,37 @@
...
+        resultsEl.appendChild(renderButton);
+
+        renderButton.onclick = function () {
+          renderWarning.style.display = 'none';
+          renderButton.style.display = 'none';
```
foolip, line 884: Since the elements are never going to get shown again, this seems cleaner:

```suggestion
          resultsEl.removeChild(renderWarning);
          resultsEl.removeChild(renderButton);
```

foolip (changes requested):

foolip (approved):

PR #1825, opened 2022-02-19
Remap event handlers

### LINKED ISSUE #1646: Remap event handlers (onfoo -> foo_event)
MDN has decided to combine data points for the events and event handlers into one, see https://github.com/mdn/browser-compat-data/pull/13595.  It seems that the `X_event` variant is preferred, so we will need to perform some remapping to accommodate.

queengooborg:
This PR fixes #1646 by remapping the names of event handlers to conform to BCD's new style, i.e. `onfoo` becomes `foo_event`.  Remapping takes place in the build script, rather than the update script, so that names displayed on the website are also remapped.


build.js
```diff
@@ -626,6 +626,9 @@ const buildIDLMemberTests = (
     }
 
     const isStatic = member.special === 'static' || iface.type === 'namespace';
+    const isEventHandler =
+      member.idlType?.type === 'attribute-type' &&
+      (member.idlType?.idlType || '').includes('EventHandler');
```
foolip, line 631: Why includes? Can it be some other type than exactly "EventHandler"?

foolip:
In addition to changing the BCD paths, I think we also need to correctly pick which interface the event should go on. For everything in `GlobalEventHandlers`, we really can't tell from the IDL. I think we will need a manually maintained mapping, and throw errors if we see event handler properties that we haven't mapped to some *_event test. Would that make sense?

build.js
queengooborg: That's correct, I've noticed some special names for event handlers for certain interfaces.

queengooborg:
I think that `GlobalEventHandlers` is actually going away now as a part of the event remapping, so I don't think we have to worry about such special cases?

foolip:
`GlobalEventHandlers` may be going away, but there are lots of *_event entries spread across many interfaces which in IDL are only on `GlobalEventHandlers`.

Putting the problem differently, how can we cover all of the existing *_event entries that find-missing-features.js lists, and also keep adding tests for new ones as the platform evolves?

queengooborg:
`GlobalEventHandlers` is one of the two remaining mixins still in BCD, so we're already flattening them out into the individual interfaces such as Document, Element, etc.  My apologies, but I don't think I'm comprehending what the issue is?  (I'm also pretty tired at the moment, so I'll give this thread another read after I've had some rest...)

foolip:
Take `onvolumechange`. How do we generate a test for api.HTMLMediaElement.volumechange_event based on that, given that `HTMLMediaElement` isn't the interface with the attribute in IDL?

queengooborg:
Ahhhhh, I see where the problem is now!  It's not the mixins that're the problem, but rather inheritance, and inherited events.

I don't have a good answer for this one, but my gut feeling is telling me that we should move the events to the interface that has the event handlers.  Even though [only media events will ever get a `volumechange` event fired](https://html.spec.whatwg.org/multipage/dom.html#:~:text=While%20these%20attributes%20apply%20to%20all%20elements%2C%20they%20are%20not%20useful%20on%20all%20elements.%20For%20example%2C%20only%20media%20elements%20will%20ever%20receive%20a%20volumechange%20event%20fired%20by%20the%20user%20agent.), the event handler is still available on all `HTMLElement` instances.

I'm also thinking this is a problem we should postpone to a later release?  As it is right now, `add-new-bcd` just wants to add the event handlers back into the data, and for the most part, this remapping _does_ work.

foolip:
> Ahhhhh, I see where the problem is now! It's not the mixins that're the problem, but rather inheritance, and inherited events.
> 
> I don't have a good answer for this one, but my gut feeling is telling me that we should move the events to the interface that has the event handlers. Even though [only media events will ever get a `volumechange` event fired](https://html.spec.whatwg.org/multipage/dom.html#:~:text=While%20these%20attributes%20apply%20to%20all%20elements%2C%20they%20are%20not%20useful%20on%20all%20elements.%20For%20example%2C%20only%20media%20elements%20will%20ever%20receive%20a%20volumechange%20event%20fired%20by%20the%20user%20agent.), the event handler is still available on all `HTMLElement` instances.

The note you link says (about the event handler properties) that "they are not useful on all elements", and I agree! I feel fairly strongly that we should not add entries for events that the browser never fires, and this means that IDL alone is not enough here.

> I'm also thinking this is a problem we should postpone to a later release? As it is right now, `add-new-bcd` just wants to add the event handlers back into the data, and for the most part, this remapping _does_ work.

Without dealing with `GlobalEventHandlers`, won't this result in tests like api.HTMLElement.volumechange_event and similar, so that add-new-bcd tries to add lots of new entries that we don't want?

If you'd like to get rid of the on* entries that add-new-bcd wants to add now, I'd say an OK interim step is to just not generate those tests.

I think what we need to generate the correct tests is a list of interface and event names, like "HTMLMediaElement" + "volumechange" which would generate a `'onvolumechange' in document.createElement('video')` test, assuming we use custom instances here.

A lot of that list can be sourced from IDL. An `onsourceopen` event handler attributes on `MediaSource` can be turned directly into "MediaSource" + "sourceopen" in the mentioned list. Perhaps we can just special-case `GlobalEventHandlers`, and require each of the on* properties there to be mapped (manually) to one or more interface+event entries on our list.

queengooborg:
Coming back to this, GlobalEventHandlers will certainly be a tough one to resolve.  Would you be down for merging this PR as-is (it does cover the majority of events properly) and then we can come back to the rest in a follow-up PR after we've figured it out in BCD?

foolip:
Yeah, I guess we can go with this for now, just mapping event handler properties to *_event entries in BCD, and just not adding all of those entries to BCD. Can you add a TODO in the code explaining how too many tests are being generated and what we'd like to do? Maybe file and link to an issue.

build.js
```diff
@@ -664,7 +670,11 @@ const buildIDLMemberTests = (
       }
     }
 
-    tests[member.name] = compileTest({
+    const name = isEventHandler ?
+      `${member.name.replace('on', '')}_event` :
```
foolip, line 674: Here, can we instead check `member.name.startsWith('on')` and then let `name = member.name.subtring(2)`?

Or we could include member.name.startsWith('on') in the definition of `isEventHandler`, so that we don't have to deal with the hypothetical case where it's not true.

foolip (changes requested):
Reviewed again with an eye to really merging it, I had a few more ideas :)

build.js
queengooborg: Hrm, this seems to have revealed a bug I wasn't expecting to see...

build.js
queengooborg: I tweaked the `replace()` call to use a small bit of regex instead that checks for `on` in the beginning of the string, which should do the above while still keeping the code to one line!

PR #1826, opened 2022-02-21
Fix find-missing-results

queengooborg:
This PR fixes the find-missing-results script.  Apparently, when performing `'15' > '4'`, the strings are not converted to numbers during comparison, but rather compared lexicographically.  This PR fixes that by replacing one of the sides of the comparison with an integer, so that IE 10 and 11 are included in the list, as well as Safari 10 or higher.

Additionally, this removes the clause to exclude Safari 6.1, since BCD no longer has 6.1 in its data.

### DIFF
--- a/find-missing-results.js
+++ b/find-missing-results.js
@@ -46,12 +46,12 @@ const generateReportMap = (allResults) => {
       if (browserKey == 'ie') {
         // Ignore super old IE releases
         result[browserKey] = result[browserKey].filter((v) =>
-          compareVersions.compare(v, '6', '>')
+          compareVersions.compare(v, '6', '>=')
         );
       } else if (browserKey == 'safari') {
         // Ignore super old Safari releases
         result[browserKey] = result[browserKey].filter((v) =>
-          compareVersions.compare(v, '4', '>')
+          compareVersions.compare(v, '4', '>=')
         );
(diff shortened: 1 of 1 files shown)


find-missing-results.js
```diff
@@ -45,12 +45,10 @@ const generateReportMap = (allResults) => {
     if (!allResults) {
       if (browserKey == 'ie') {
         // Ignore super old IE releases
-        result[browserKey] = result[browserKey].filter((v) => v >= '6');
+        result[browserKey] = result[browserKey].filter((v) => v >= 6);
```
foolip, line 48: Can you use `compareVersions` here so that the string&rarr;number conversion isn't implicit?

foolip (changes requested):

find-missing-results.js
```diff
@@ -45,11 +45,13 @@ const generateReportMap = (allResults) => {
     if (!allResults) {
       if (browserKey == 'ie') {
         // Ignore super old IE releases
-        result[browserKey] = result[browserKey].filter((v) => v >= '6');
+        result[browserKey] = result[browserKey].filter((v) =>
+          compareVersions.compare(v, '6', '>')
```
foolip, line 49: Did you intend to change `>=` to `>`?

PR #1903, opened 2022-04-14
Set to production by default if NODE_ENV = production

queengooborg:
This PR reverses the preference of staging vs. production when `NODE_ENV` is production.  Currently, the website is only in production mode when the `GAE_VERSION` is set to `production`.  However, this essentially means that with our current code, we can only launch a production deployment on Google App Engine.  This PR reverses the preference, so that production is the default when `NODE_ENV` is `production`. ~~, and staging deployment is on when `GAE_VERSION` is `staging`.~~

~~@foolip, can you confirm the staging version has `GAE_VERSION=staging`?~~

Edit: since the staging version was removed from GAE, this PR has now been adjusted so that it is only based upon `NODE_ENV` and has no logic for `GAE_VERSION`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -40,7 +40,7 @@ const storage = getStorage();
 
 /* c8 ignore start */
 const appVersion =
-  process.env.NODE_ENV === 'production' && process.env.GAE_VERSION !== 'staging' ?
+  process.env.NODE_ENV === 'production' ?
     (await fs.readJson(new URL('./package.json', import.meta.url))).version :
     'Dev';
 
--- a/app.js
+++ b/app.js
@@ -40,7 +40,7 @@ const storage = getStorage();
 
 /* c8 ignore start */
 const appVersion =
-  process.env.GAE_VERSION === 'production' ?
+  process.env.NODE_ENV === 'production' && process.env.GAE_VERSION !== 'staging' ?
     (await fs.readJson(new URL('./package.json', import.meta.url))).version :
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -40,7 +40,7 @@ const storage = getStorage();
 
 /* c8 ignore start */
 const appVersion =
-  process.env.GAE_VERSION === 'production' ?
+  process.env.NODE_ENV === 'production' ?
```
foolip, line 43: Is there any particular reason we can't just let this code read package.json unconditionally? Without a staging environment there's just local and testing and we won't confuse ourselves there I think.

queengooborg: The rest of our code relies on appVersion being set to `Dev` to determine whether or not we're running a development build, so that it may report as such.  For example, if a PR is submitted from the dev version, the PR message will report as such to indicate we probably shouldn't merge it.  On that note, technically speaking, the version reported by `package.json` won't be accurate on dev builds since we update the version number just before a release.

I had the idea of using the shortened commit hash as the version number when running a development build, but I didn't want to worry about implementing such logic at the time.  How would that sound to you, though?

foolip: I see, so we do use this value for stuff beyond the UI. Let's keep this simplification to the current scope then.

Using `git describe` would make sense I think, but in a separate PR if so.

foolip (approved):

PR #1912, opened 2022-04-18
Use 'git describe' to generate version number for dev version

queengooborg:
This PR updates the app to use `git describe` to create a version number for the development version.

### DIFF
--- a/app.js
+++ b/app.js
@@ -35,6 +35,7 @@ import parseResults from './results.js';
 import {getStorage} from './storage.js';
 import {parseUA} from './ua-parser.js';
 import Tests from './tests.js';
+import {exec} from './scripts.js';
 
 const storage = getStorage();
 
@@ -42,7 +43,7 @@ const storage = getStorage();
 const appVersion =
   process.env.NODE_ENV === 'production' ?
     (await fs.readJson(new URL('./package.json', import.meta.url))).version :
-    'Dev';
+    'Dev-' + String(exec('git describe --tags')).substr(1);
 
 const secrets = await fs.readJson(
   new URL(
--- a/app.js
+++ b/app.js
@@ -40,18 +40,22 @@ import {exec} from './scripts.js';
 const storage = getStorage();
 
 /* c8 ignore start */
-let appVersion;
-if (process.env.NODE_ENV === 'production') {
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -35,14 +35,23 @@ import parseResults from './results.js';
...
-  process.env.NODE_ENV === 'production' ?
-    (await fs.readJson(new URL('./package.json', import.meta.url))).version :
-    'Dev';
+let appVersion;
+if (process.env.NODE_ENV === 'production') {
+  appVersion = (await fs.readJson(new URL('./package.json', import.meta.url)))
+    .version;
+} else {
+  try {
+    appVersion = 'Dev-' + String(exec('git describe --tags')).substr(1);
```
foolip, line 49: The `substr(1)` bit is assuming that that the tag starts with "v" and will result in something broken if that's not the case. How about we instead prepend a "v" to the version read from package.json, and make no assumptions about what the tag found by `git describe --tags` is?

See next comment for what to do about "Dev".

exporter.js
```diff
@@ -73,7 +73,7 @@ const createBody = (meta) => {
     }` +
     `\nHash Digest: ${meta.digest}` +
     `\nTest URLs: ${meta.urls.join(', ')}` +
-    (meta.version == 'Dev' ?
+    (meta.version.startsWith('Dev') ?
```
foolip, line 76:
```suggestion
    (meta.version.includes('-') ?
```

app.js
```diff
@@ -35,14 +35,23 @@ import parseResults from './results.js';
...
+  } catch (e) {
+    // If anything happens, i.e. git isn't installed, ignore
```
foolip, line 51: I think let's instead fall back to package.json. So if `NODE_ENV !== 'production'` we try using `git describe --tags`, and if we don't get and return a version, we read it from package.json.

foolip (changes requested):

exporter.js
queengooborg: I'm skeptical about using a hyphen to indicate a dev version or not, but I'm down for using `includes()` instead!

foolip:
Hmm, I made the tests fail. Not sure why.

queengooborg:
Ah, the tests still use just "Dev" as the version in the dummy results, so that'll need to be tweaked.

PR #1934, opened 2022-04-21
Use a single storage bucket with app version as path prefix

foolip:
The main purpose of this is to stop using GAE_VERSION, to allow it to be
anything and make it easy to deploy multiple versions in parallel.

Use the app version as a path prefix in the single bucket to clearly
separate data from different versions, and remove the need for a staging
bucket. Collisions were already unlikely, but sessions for all
deployments would still be mixed in a single bucket.

While changing things, also add /sessions/ to the path of temporary
results storage, since /files/ is already used for exported reports.

### DIFF
--- a/app.js
+++ b/app.js
@@ -37,8 +37,6 @@ import {parseUA} from './ua-parser.js';
 import Tests from './tests.js';
 import {exec} from './scripts.js';
 
-const storage = getStorage();
-
 /* c8 ignore start */
 const getAppVersion = async () => {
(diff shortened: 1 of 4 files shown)


queengooborg (approved):

foolip:
I've tried this with https://pr1934-dot-mdn-bcd-collector.ue.r.appspot.com/ and it works end-to-end, so I'll prepare a v6 release now.

PR #1960, opened 2022-04-26
Fix api.CSSKeyframesRule, which is currently invalid JS

gsnedders:
This contains a new line within a JS string, which is syntactically invalid, and results in every browser giving null for this test.


queengooborg (approved):
Thanks for fixing this up!

foolip:
@queengooborg can you check if there are more cases like this introduced at the same time?

PR #1983, opened 2022-05-06
Return null (not false) for unavailable reusable instances

foolip:
This won't change any test results, but makes more sense.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -24,7 +24,7 @@ api:
       src: |-
         var constructor = window.AudioContext || window.webkitAudioContext;
         if (!constructor) {
-          return false;
+          return null;
         }
         return new constructor();
     createStyleSheet:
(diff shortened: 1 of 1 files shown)


queengooborg:
Are you sure that _none_ of the test results change because of this?  I would expect this to produce a bunch more `nulls` from `cannot use "in" for false/null` errors?

foolip:
@queengooborg since this code is wrapped in a function to create the reusable instance and isn't just inlined into the test, the `return false` statements don't result in the overall test returning false. If I'm reading things right, it should just result in `reusableInstances.audioContext` and similar being set to false. Example from http://mdn-bcd-collector.appspot.com/tests/api/AudioContext/close:

```js
reusableInstances.audioContext = (function () {
  var constructor = window.AudioContext || window.webkitAudioContext;
  if (!constructor) {
    return false;
  }
  return new constructor();
})()

(function() {
  var instance = reusableInstances.audioContext;
  return 'close' in instance;
})();
```

`'close' in false` and `'close' in null` are both broken. If we wanted this code to return false if no reusable instance can be created, something needs to be added.

PR #1986, opened 2022-05-09
Remove custom tests for CSSConditionRule and CSSGroupingRule

foolip:
These are abstract base classes that were added to the prototype chain
after some of the interfaces that inherit from them. Creating instances
for these leads to detected support before the interfaces were actually
supported.

Simple prototype testing is appropriate in this case, although it's
possible some browser supported the interfaces without exposing the
interface objects. Regardless of how this is testing, it will need to be
manually scrutinized when updating the data.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -396,10 +396,6 @@ api:
       - cryptoKey
     __base: |-
       var promise = reusableInstances.cryptoKey;
-  CSSConditionRule:
-    __resources:
-      - createStyleSheet
-    __base: <%api.CSSMediaRule:instance%>
   CSSCounterStyleRule:
     __resources:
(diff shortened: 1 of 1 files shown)


foolip:
Ping @queengooborg

queengooborg:
Removing the tests feels odd to me, especially since these interface were supported before prototype chain exposure.  Perhaps instead, we could add name checking?

foolip:
@queengooborg do we check the name in some other custom tests that I could copy from? Note that this would be the name of an intermediate prototype, not the "leaf" interface.

foolip:
This was made obsolete by https://github.com/foolip/mdn-bcd-collector/pull/2718.

Issue #1520, opened 2021-09-23
MediaStream tests depend on navigator.mediaDevices.getUserMedia()


foolip:
The custom tests for `MediaStream` depend on `navigator.mediaDevices.getUserMedia()`:

https://github.com/foolip/mdn-bcd-collector/blob/45c1a30614f08f7131a49bfa1d72d7b94ca81b00/custom-tests.yaml#L1195-L1199

This isn't good, because `navigator.mediaDevices.getUserMedia()` is much newer than `MediaStream`. It's especially problematic because there's a generated `if (!mediaDevices) { return false; }` in the tests, which means "definitely not supported".

This also affects `MediaStreamTrack` tests.

cc @queengooborg

queengooborg:
We have an issue open about adding the callback-based version into the collector, see #1345.  Let me see what we can do for this!

foolip:
Looks like I noticed the same problem in https://github.com/mdn/browser-compat-data/pull/11718#issuecomment-886199688 but didn't file an issue then.

foolip:
Running across this again today, it's not possible to detect support for `MediaStreamTrack`'s `stop()` method correctly because of this.

PR #1991, opened 2022-05-10
Fall back to legacy getUserMedia for MediaStream instance

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/1520.

### DIFF
--- a/custom-tests.yaml
+++ b/custom-tests.yaml
@@ -1148,8 +1148,7 @@ api:
     __base: var instance = createImageBitmap(document.getElementById('resource-image-black'));
   ImageCapture:
     __base: |-
-      <%api.MediaDevices:mediaDevices%>
-      var stream = mediaDevices.getUserMedia({video: true});
+      var stream = navigator.mediaDevices.getUserMedia({video: true});
       stream.then(function() {});
       var promise = stream.then(function(s) {
         var tracks = s.getVideoTracks();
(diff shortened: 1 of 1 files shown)


custom-tests.yaml
```diff
@@ -1233,9 +1231,18 @@ api:
     isTypeSupported: return 'isTypeSupported' in MediaSource;
   MediaStream:
     __base: |-
-      <%api.MediaDevices:mediaDevices%>
-      var promise = mediaDevices.getUserMedia({audio: true});
-      promise.then(function() {});
+      // This string makes this a callback test: callback(
```
foolip, line 1234: @queengooborg this isn't very nice. Do you have any thoughts on how to write the detection instead? Can we more explicitly mark tests as callback tests?

foolip: Ping @queengooborg

queengooborg: Actually, this string isn't needed!  As long as the string `callback` is present, then we're good to go, no open parenthesis needed!

foolip: Isn't this the code that determines if it's a callback test?

https://github.com/foolip/mdn-bcd-collector/blob/52df4d5551217feb9f0b1d42b18228caa25a83ef/build.js#L90

There's an open paren there...

queengooborg: Oh I was looking at the functions above (we should probably split it off into its own function), hmm...  We could perhaps turn it into a regex check to find a comma or right parenthesis afterwards as well?

foolip: Can we add the test type as a bit of info in the YAML perhaps, to serve at least as an override for the detected type?

foolip: So with https://github.com/foolip/mdn-bcd-collector/pull/2253 I can drop the comment here.

foolip:
Closing in favor of https://github.com/foolip/mdn-bcd-collector/pull/2589.

PR #2009, opened 2022-05-17
Revert "Randomize selection of Selenium host"

queengooborg:
Reverts foolip/mdn-bcd-collector#1919.  Due to the variation of results between CTs, it's not a good idea to mix results if we can help it.

### DIFF
--- a/selenium.js
+++ b/selenium.js
@@ -99,9 +99,6 @@ const ignore = {
   }
 };
 
-// Sort array randomly
-const shuffle = () => Math.random() - 0.5;
-
 const prettyName = (browser, version, os) => {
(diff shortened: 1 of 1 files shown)


selenium.js
```diff
@@ -99,9 +99,6 @@ const ignore = {
   }
 };
 
-// Sort array randomly
-const shuffle = () => Math.random() - 0.5;
```
foolip, line 103: https://www.robweir.com/blog/2010/02/microsoft-random-browser-ballot.html might be amusing reading, this method of shuffling is imperfect and has been a high profile "browser choice" bug in the past!

PR #2054, opened 2022-06-06
Allow for static members of the same name as non-static members

queengooborg:
This PR allows the IDL to contain static members that use the same name as non-static members, as per the [WebIDL standard](https://webidl.spec.whatwg.org/\#idl-operations:\~:text\=Note%3A%20The%20identifier%20of%20a%20static%20operation%20can%20be%20the%20same%20as%20the%20identifier%20of%20a%20regular%20operation%20defined%20on%20the%20same%20interface.).  This is required to land #2049.

Note: while this logic can theoretically be simplified to `member.name + member.special`, I wanted to limit the scope to just what's allowed in the spec.

### DIFF
--- a/build.js
+++ b/build.js
@@ -266,12 +266,12 @@ const compileTest = (test) => {
 const mergeMembers = (target, source) => {
   // Check for duplicate members across partials/mixins.
   const targetMembers = new Set(
-    target.members.map((m) => (m.name + m.special === 'static' ? 'static' : ''))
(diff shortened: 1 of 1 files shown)


foolip:
@queengooborg this changed http://localhost:8080/tests/api/Response/json to be `"json" in Response` instead of:

```js
(function() {
  var instance = new Response('');
  return 'json' in instance;
})();
```

This is wrong, since the existing entry in BCD is about the instance method.

I've also filed https://github.com/mdn/browser-compat-data/issues/16613 about this clash in BCD.

PR #2061, opened 2022-06-09
Don't generate a test for static Response.json()

foolip:
This fixes an accidentally changed test for this PR:
https://github.com/foolip/mdn-bcd-collector/pull/2054

See https://github.com/mdn/browser-compat-data/issues/16613.


queengooborg:
Thanks @foolip!  I opted to write #2076 instead to turn this into less of a special case, should more of these pop up.

PR #2076, opened 2022-06-19
Don't include static members with the same name as non-static members

queengooborg:
This PR is a follow-up to #2054.  Instead of simply ignoring static members with the same name as non-static members, this PR adjusts the logic to ensure they aren't included in the merged member output at all.  Closes #2061.


foolip:
This didn't work, https://mdn-bcd-collector.appspot.com/tests/api/Response/json is still this test for the static function:

```js
"Response" in self && "json" in Response
```

PR #2179, opened 2022-07-24
Clearly state support vs. true/false/null

queengooborg:
This PR replaces the `true/false/null` display with `Supported/No Support/Support Unknown` to make it clearer to end users whether a feature is supported or not.


foolip:
Cool, thanks!

PR #2202, opened 2022-08-03
Track all individual minor versions of Safari 15+

queengooborg:
This PR updates the ua-parser script to track all Safari 15+ minor versions as individual releases, rather than mapping down (as needed for Samsung Internet and older Safari releases).

### DIFF
--- a/ua-parser.js
+++ b/ua-parser.js
@@ -106,25 +106,17 @@ const parseUA = (userAgent, browsers) => {
   // with this, find the pair of versions in |versions| that sandwiches
   // |version|, and use the first of this pair. For example, given |version|
   // "10.1" and |versions| entries "10.0" and "10.2", return "10.0".
-  if (
-    data.browser.id.startsWith('safari') &&
-    compareVersions.compare(data.version, '15', '>=')
-  ) {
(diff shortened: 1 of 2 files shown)


foolip (commented):
Can you add some tests for this? I'd expect the current code to work if BCD has the individual 15.x releases, but that there will be a problem with 15.6 if only 15.5 is in BCD currently. But this problem isn't explicitly handled here...

Ultimately, which cases will this change?

foolip:
Thinking about this some more, I don't think there was actually a change with regards to minor versions at 15+, only the pace. Both before at after, the minor version is often significant for Safari, but we might have minor versions that we'll skip in BCD if there were no web exposed changes.

So some examples would help.

queengooborg:
> I don't think there was actually a change with regards to minor versions at 15+, only the pace

I meant to say "Safari iOS", since there would be a few minor releases of iOS in between Safari minor releases.  For Safari Desktop, every minor version is just as important as every major version, especially with the more frequent updates now!

foolip:
https://github.com/foolip/mdn-bcd-collector/pull/2208 is more like what I'd prefer here. 15+ aren't special, right?

PR #2205, opened 2022-08-03
Add overrides for streaming webassembly features in Chrome 60

foolip:


foolip:
This case is weird, I haven't quite gotten to the bottom of it. The feature was enabled for 61:
https://storage.googleapis.com/chromium-find-releases-static/e16.html#e1628530d4237e822cd2c3125ab476ea87281d30

But it seems like the methods were exposed in 60 but didn't work. I haven't looked for the commits where that mishap occurred.

PR #2208, opened 2022-08-03
Improve handling of future (not in BCD) releases

foolip:


ua-parser.js
```diff
@@ -9,6 +9,11 @@
 import compareVersions from 'compare-versions';
 import uaParser from 'ua-parser-js';
 
+const getMajorVersion = (version) => {
+  const [major] = version.split('.');
+  return major;
```
foolip, line 14:
```suggestion
  return version.split('.')[0];
```

PR #2212, opened 2022-08-05
Don't generate a test for static Response.json()

foolip:
This fixes an accidentally changed test for this PR:
https://github.com/foolip/mdn-bcd-collector/pull/2054

See https://github.com/mdn/browser-compat-data/issues/16613.

### DIFF
--- a/build.js
+++ b/build.js
@@ -643,6 +643,13 @@ const buildIDLMemberTests = (
   const handledMemberNames = new Set();
 
   for (const member of members) {
+    // Skip static Response.json() so it doesn't clobber the test for
+    // Response.prototype.json, see
+    // https://github.com/mdn/browser-compat-data/issues/16613.
+    if (iface.name === 'Response' && member.special === 'static' && member.name === 'json') {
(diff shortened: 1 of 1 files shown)


foolip:
Dupe of https://github.com/foolip/mdn-bcd-collector/pull/2061.

foolip:
So this is a dupe of https://github.com/foolip/mdn-bcd-collector/pull/2061, but I can't reopen that now, probably because the branch name is the same.

foolip:
We also need to adapt the alternative fix in https://github.com/foolip/mdn-bcd-collector/pull/2076 somehow. Trying a revert in https://github.com/foolip/mdn-bcd-collector/pull/2213 to see if it's needed to pass any tests.

queengooborg:
I don't like the idea of special-casing `Response.json`, and I feel that we should operate under the assumption that more cases like this will appear.  I've opened https://github.com/foolip/mdn-bcd-collector/pull/2218 to fix the logic from #2076.

PR #2213, opened 2022-08-05
Revert "Don't include static members with the same name as non-static members"

foolip:
Reverts foolip/mdn-bcd-collector#2076

### DIFF
--- a/build.js
+++ b/build.js
@@ -267,27 +267,20 @@ const compileTest = (test) => {
 
 const mergeMembers = (target, source) => {
   // Check for duplicate members across partials/mixins.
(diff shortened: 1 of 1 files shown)


foolip:
Since all the tests pass and this didn't fix the `Response.json()` issue, I guess reverted this is a no-op and we should do it.

queengooborg:
I found what was going on with the code -- it seems as though the `targetMembers` object was treated as a different object than the one in the array, so using `indexOf()` didn't work.  I've opened #2218 to fix the logic!

queengooborg:
Since the logic's now fixed (and it is now confirmed to be a working fix), I'm going to go ahead and close this reversion PR.

PR #2239, opened 2022-08-16
Add overrides for api.FontFaceSetLoadEvent

queengooborg:
This PR adds overrides for `api.FontFaceSetLoadEvent` in response to https://github.com/mdn/browser-compat-data/pull/17365.

### DIFF
--- a/overrides.json
+++ b/overrides.json
@@ -19,6 +19,6 @@
     false
   ],
   "https://github.com/mdn/browser-compat-data/pull/17365#issuecomment-1216459968",
-  ["api.FontFaceSetLoadEvent", "chrome", "35", true],
-  ["api.FontFaceSetLoadEvent.fontfaces", "chrome", "35", true]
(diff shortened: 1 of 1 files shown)


overrides.json
```diff
@@ -17,5 +17,8 @@
     "chrome",
     "60",
     false
-  ]
+  ],
+  "https://github.com/mdn/browser-compat-data/pull/17365#issuecomment-1216459968",
+  ["api.FontFaceSetLoadEvent", "chrome", "35", true],
```
foolip, line 22: Will this override only change what the results in Chrome 35 are, causing a flip-flopping inferred statement?

queengooborg: If anything, this will create a complex inferred statement, and thus end up not updating the data.  Maybe we should allow for setting a range as the version (ex. `aa-bb`) so we don't have to duplicate this statement so many times?

PR #2264, opened 2022-08-24
Disable the unsecure warning for now

queengooborg:
The unsecure warning is still displaying even when it shouldn't.  I have a feeling this may be an issue around GAE.  CC @foolip


foolip:
@queengooborg can you file an issue? I'll never get back to this as just a notification :)

PR #2277, opened 2022-08-29
Correct reference to project license

jugglinmike:
This patch was generated from the following command:

    $ git ls-files -z | xargs --null sed -i 's/See LICENSE\.txt/See the LICENSE file/g'

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See [DESIGN.md](./DESIGN.md) for details of how this service works.
 
 ## Copyright
 
-This project is © 2020-2022 Google LLC, Mozilla Corporation and Gooborg Studios, Apache License 2.0. See LICENSE.txt for more details.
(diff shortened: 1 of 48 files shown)


foolip:
Renaming LICENSE to LICENSE.txt would also do the trick. What do other MDN repos do? There's a chance this repo will be transferred, so consistency can help decide.

jugglinmike:
Good point! [browser-compat-data](https://github.com/mdn/browser-compat-data) uses "LICENSE" consistently:

    $ git log --oneline -n1
    76926f4a2 (HEAD -> main, upstream/main, upstream/HEAD) Make note about return value of prompt() (#17541)
    $ ls LICENSE*
    LICENSE
    $ git grep LICENSE.txt
    $ git grep 'LICENSE file' | wc -l
    75

queengooborg (approved):
Thank you, LGTM!

Issue #2261, opened 2022-08-22
Potential inconsistency in the handling of the "mirror" value


jugglinmike:
Situation: a browser's "support" value is set to `"mirror"` and the `update-bcd` script encounters test results which differ from those of the "upstream" browser

Actual behavior: The script modifies the "support" value, and it does so only in some cases.

Expected behavior: The script should consistently update the value (or consistently fail since the input contradicts the expected correlation).

Here's a table that describes the behavior using a set of fake API tests designed to highlight this behavior. It uses the "Oculus" browser to observe if/how the script updates "support" under a variety of conditions:

Chrome support                      | Chrome result | Oculus support (before) | Oculus result | Oculus support (after)
------------------------------------|---------------|-------------------------|---------------|-----------------------
{"version_added": 3}                | true          | "mirror"                | true          | "mirror"
{"version_added": 3}                | true          | "mirror"                | false         | "mirror"
{"version_added": 3}                | true          | "mirror"                | null          | "mirror"
{"version_added": 3}                | false         | "mirror"                | true          | "mirror"
{"version_added": 3}                | false         | "mirror"                | false         | "mirror"
{"version_added": 3}                | false         | "mirror"                | null          | "mirror"
{"version_added": 3}                | null          | "mirror"                | true          | "mirror"
{"version_added": 3}                | null          | "mirror"                | false         | "mirror"
{"version_added": 3}                | null          | "mirror"                | null          | "mirror"
{"version_added": 3, "flags": [{}]} | true          | "mirror"                | true          | [{"version_added": "≤22"}, "mirror"]
{"version_added": 3, "flags": [{}]} | true          | "mirror"                | false         | "mirror"
{"version_added": 3, "flags": [{}]} | true          | "mirror"                | null          | "mirror"
{"version_added": 3, "flags": [{}]} | false         | "mirror"                | true          | [{"version_added": "≤22"}, "mirror"]
{"version_added": 3, "flags": [{}]} | false         | "mirror"                | false         | "mirror"
{"version_added": 3, "flags": [{}]} | false         | "mirror"                | null          | "mirror"
{"version_added": 3, "flags": [{}]} | null          | "mirror"                | true          | [{"version_added": "≤22"}, "mirror"]
{"version_added": 3, "flags": [{}]} | null          | "mirror"                | false         | "mirror"
{"version_added": 3, "flags": [{}]} | null          | "mirror"                | null          | "mirror"
[...]

Issue #571, opened 2020-09-30
update-bcd: various flaws with the design


foolip:
At a high level, update-bcd takes a one or more [test reports](https://github.com/foolip/mdn-bcd-results) as input, and modifies a copy of BCD in place to match the test results.

The test results are potentially conflicting, for multiple reasons:
- There can be multiple tests for a single BCD entry, using different testing methodology (https://github.com/foolip/mdn-bcd-collector/issues/523)
- Multiple browser+platform combinations can map to a single BCD browser release (for example, Chrome on macOS and Windows both map to `chrome`)
- Multiple browser releases can map to a single BCD release (see https://github.com/foolip/mdn-bcd-collector/pull/565)

The current design of the update-bcd script is:
- Build a "support matrix" mapping a BCD entry + browser release to a support status (true/false/null) and sometimes a prefix (this also resolves contradictory results for a single BCD entry)
- For every BCD entry (in `api/**.json` and `css/**.json`) for which we have data in the support matrix:
  - Infer a BCD-style support statement from the per-version data in the support matrix. Where we have incomplete information, ranged (≤) versions are used.
  - If the inferred support statement isn't simple (a single object) give up and do nothing.
  - If the BCD support statement had only `prefix`/`alternative_name`/`flags` entries, add our inferred statement
  - Otherwise, there was an existing simple statement, which we update using the inferred statement

There are a bunch of problems here:
- We don't correctly handle `prefix` and `alternative_name`, or at least not having multiple variants, as the support matrix squashes this information together too early.
- We do nothing if the inferred support statement is too complicated (so for example, we don't update BCD if we've found a add-remove-and-add-again situation, regardless of the existing state of BCD)
- We can never add a support statement if there's already a simple statement, only update it.

Not all of these are very serious, and I don't think any of them cause us to suggest incorrect updates. However, we're leaving some updates on the table by bailing in the situations.

foolip:
One feature of an improved design that I'd like to see is that we have an explicit test for contradictions between our results for a specific browser release and a BCD-style support statement. Given that, we could apply that test before and after we've updated, and much more easily see what contradictions weren't resolved by the updates. Maybe we'll see that there aren't many and we'd be better off resolving those contradictions manually.

foolip:
Hmm, I see that actually we can't deal with `prefix` or `alternative_name` at all, because we never look at the `info` object in the results where the prefix would be.

foolip:
[...]

PR #2280, opened 2022-08-30
Correct support for mirroring

jugglinmike:
Hi there! I'm interested in contributing to BCD with authentic test results for the Meta Quest browser, so I've been working with the "mirror" feature ([as introduced to BCD here](https://github.com/mdn/browser-compat-data/issues/15083) and integrated to this project in gh-2039). I filed gh-2261 to report my initial experience, but even following [an expeditious fix by @queengooborg](https://github.com/foolip/mdn-bcd-collector/pull/2265), I suspect there may still be some bugs.

To facilitate this discussion, I've written nine unit tests for the mirroring behavior implemented by [the `update-bcd` script](https://github.com/foolip/mdn-bcd-collector/blob/707497c83db2c31e9727eaa12cde6b3e52f69052/update-bcd.ts). The first commit in this pull request introduces those tests, including "expected" values dictated by the current behavior of the application. However, as mentioned above, some of the current behavior seems incorrect to me. The second commit modifies the "expected" values according to how I believe the feature was conceived to behave. This commit does *not* modify the application logic (I'd like to verify my interpretation with the maintainers before attempting to implement a solution), so the new tests fail with the second commit applied.

(I'm also happy to iterate on the organization of the test code, ideally in parallel to the discussion about the expected behavior of the application code. I've tried to strike a balance between concision and legibility--a never-ending struggle in test design. Also: although the existing tests each tend to validate a bunch of behaviors, I tried to express the conditions here with a series of focused cases. That said, I wasn't able to completely avoid dependencies on the surrounding state because browser data, although parameterized in this project, is [hard-coded into the BCD project](https://github.com/mdn/browser-compat-data/blob/76926f4a2ffea3c0a8a5bdc8214fba4200d8e1ab/scripts/release/mirror.ts#L15-L16).)


unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+          upstream: 'chrome',
+          releases: {86: {}}
+        }
+      };
+
+      const bcdFromSupport = (support) => ({
+        api: {FakeInterface: {__compat: {support}}}
+      });
+
+      const mirroringCase = ({support, downstreamResult}) => {
```
foolip, line 879: This is a useful helper. Can you comment that it produces a report for Chrome Android 86? This is important for understanding the tests that use this, and his hidden in the UA string at the bottom.

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+              chrome: {version_added: '85'},
+              chrome_android: 'mirror'
+            },
+            downstreamResult: true
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: '85'},
+              chrome_android: {version_added: '86'}
```
foolip, line 915: Without results from Chrome Android 85, I don't think this is a valid assumption. The updater script should have concluded "support in Chrome Android 86, earlier support unknown" and not see a contradiction in need of fixing by updating the data.

The updater code isn't really structured in a way that makes this easy to fix however, see https://github.com/foolip/mdn-bcd-collector/issues/571

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+              chrome: {version_added: '85'},
+              chrome_android: 'mirror'
+            },
+            downstreamResult: false
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: '85'},
+              chrome_android: false
```
foolip, line 932: Agreed.

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+              chrome: {version_added: '85'},
+              chrome_android: 'mirror'
+            },
+            downstreamResult: null
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: '85'},
+              chrome_android: 'mirror'
```
foolip, line 949: Yep, there was no contradiction, so we should leave it as whatever it was.

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+            },
+            downstreamResult: true
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: '85', flags: [{}]},
+              chrome_android: [
+                {version_added: '86'},
+                {flags: [{}], version_added: '85'}
```
foolip, line 970: The guideline on this was revised in https://github.com/mdn/browser-compat-data/pull/16637, and we should just omit the flag data. If this is the current behavior a TODO somewhere is fine though.

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,205 @@ describe('BCD updater', () => {
...
+              chrome: {version_added: false},
+              chrome_android: 'mirror'
+            },
+            downstreamResult: true
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: false},
+              chrome_android: {version_added: '86'}
```
foolip, line 1024: Again this is not a valid inference. The code should be adding a range here, right? (A human then needs to fix it, because the range isn't valid in BCD.)

foolip (commented):
Thanks Mike, and nice to see you here! I haven't looked at the current behavior after @queengooborg made some changes, but I've commented on what I think _should_ be the behavior.

Note that the ambiguity when having just a single browser version's results applies outside of mirroring too.

unittest/unit/update-bcd.ts
jugglinmike: Certainly. Your comment also makes it clear that the UA string was a magic constant, so I've stored it in an appropriately-named variable to further underscore the intent of this code.

unittest/unit/update-bcd.ts
jugglinmike: > Without results from Chrome Android 85, I don't think this is a valid assumption. The updater script should have concluded "support in Chrome Android 86, earlier support unknown" and not see a contradiction in need of fixing by updating the data.

I think I understand. The expected value for this test should be "mirror", right?

> The updater code isn't really structured in a way that makes this easy to fix however, see #571

The code currently produces the value "mirror" for this case (as per the first commit on this branch), so if that matches the expectation, then there's nothing to fix here.

unittest/unit/update-bcd.ts
jugglinmike: I'm happy to leave this for some future patch, but I don't know how to write the comment, yet. Should the flag data be persisted in the Chrome support? For example:

```
     chrome: {version_added: '85', flags: [{}]},
     chrome_android: {version_added: '85'}
```

Or should it be removed there, too? For example:

```
     chrome: {version_added: '85'},
     chrome_android: {version_added: '85'}
```

In the latter case, should the script persist the `"mirror"` value for Chrome Android? For example:

```
     chrome: {version_added: '85'},
     chrome_android: 'mirror'
```

unittest/unit/update-bcd.ts
jugglinmike: Yes! Thanks! In the `version_added` property in surrounding tests, I'm seeing ranges like `'0> ≤86'` and like `'≤86'`. [This project's description of range format](https://github.com/foolip/mdn-bcd-collector/blob/cec271d457776a375939a1b674c69fec154ea1d1/README.md#custom-ranged-version-format) doesn't seem to apply to the `0>` part. Based solely on a mathematical background, these seem equivalent. Is that right? If so, which should this test use?

jugglinmike (commented):
Thanks, Philip, nice to be working with you again! It's been too long since I've had the pleasure of a foolip review :)

unittest/unit/update-bcd.ts
foolip: Yes, the way I'd like the updater to work is that when we have incomplete information that doesn't contradict the existing data, it is left alone. (An early goal of being able to derive everything in BCD from test results has gone out the window by now, but might explain some choices in the design.)

What you might want to do is to tweak these tests to actually work with a false-to-true transition, to test the behavior I think you were originally going for here.

The case of incomplete information can be tested much more lightly, maybe just one "no change made" test, and perhaps one for the case  where we can say the current data is wrong, but not exactly what is right. Such tests need not involve mirroring though.

foolip: How to deal with incomplete information is indeed beyond the scope of fixing mirroring, but the test expectations you've written seem to suggest you'd like to change it, unless I'm wrong about current behavior.

The way I'd suggest approaching mirroring is:
- Mirror the data from upstream
- Do everything the same way as if mirroring didn't exist
- Optionally at the end, detect if the new data is the same as mirroring would produce and if so replace with "mirror". (This isn't super important because one can always run the mirror migration script in BCD later to fix it.)

My comments are about how the "do everything" bit should work, since the tests poke at that. The "transition" I'm talking about is not changing a value in BCD, but the logic in `inferSupportStatements`. If we have results for Chrome Android 85 and 86 and observe a false&rarr;true transition in test results between the two, then we can infer `{version_added: "86"}`. But if we have results from *only* 86, then the correct inference is `{version_added: "≤86"}`. That wont pass the BCD lint, but it does force a human to look into the uncertainty.

If you generate results for both 85 and 86 you can deal only with the simpler case, which I suspect is the more important one to get fixed.

unittest/unit/update-bcd.ts
foolip: This is also a situation that's independent of mirroring in principle. Whether we are updating Chrome or Chrome Android data, if the current data is just flags, and we're adding an entry for unconditional support, then we should just strip out any flag data. So perhaps:

```
// TODO: Remove flag data when adding enabled-by-default data. See https://github.com/mdn/browser-compat-data/pull/16637.
```

unittest/unit/update-bcd.ts
foolip: That range format is a bit odd and basically is to shout "human, look at me!"

But the idea is that if we've confirmed no support in release A, and having support in version C, but no results from version B in between them, then we generate `A> ≤C` to record the information that we had, instead of the simpler `≤C`. I guess there's just no simplification done when A is the first release, since we haven't run into that with current browsers.

If you use mock data that doesn't result in these ranges that'd be fine, but if you'd like to exercise it, just use whatever is currently produced, I don't think it's wrong, just hard to parse.

unittest/unit/update-bcd.ts
jugglinmike: My goal is to contribute data about the Meta Quest browser (formerly Oculus Quest). I'm focused on the mirroring behavior of `update-bcd` script because it's my understanding that (due to [this patch](https://github.com/mdn/browser-compat-data/pull/16895)), I'll be relying heavily on that specific functionality. As I understand it, the incomplete information in this case would be compatibility data for prior releases of the Meta Quest browser.

That's why all of these tests describe conditions where the "downstream" support value is set to `"mirror"` and why none include prior support values downstream. I thought I was covering all the conditions for my use case exhaustively, but I'm getting the sense that either my understanding of the use case is inaccurate or that the underlying problem is larger than what can be observed (or addressed) by my use case alone.

Does that match your understanding?

unittest/unit/update-bcd.ts
jugglinmike: Currently (as per the first commit on this branch), the script leaves the value as `"mirror"`. I'll just choose one for now and adjust if I find that the simplest fix favors the other.

unittest/unit/update-bcd.ts
jugglinmike: Thanks!

unittest/unit/update-bcd.ts
jugglinmike: I'm just now realizing that this is a typo. I think it should instead be expressed as `chrome_android: {version_added: false}`, right?

unittest/unit/update-bcd.ts
```diff
@@ -861,5 +861,216 @@ describe('BCD updater', () => {
...
+              chrome: {version_added: '85', flags: [{}]},
+              chrome_android: 'mirror'
+            },
+            downstreamResult: false
+          });
+          assert.deepEqual(
+            actual,
+            bcdFromSupport({
+              chrome: {version_added: '85', flags: [{}]},
+              chrome_android: false
```
jugglinmike, line 999: Also here: a "support statement" is either `"mirror"` or a "simple support statement", and a "simple support statement" must be an Object value. So this expected value ought to be expressed with `chrome_android: {version_added: false}`.

unittest/unit/update-bcd.ts
foolip: I think I might be getting you entangled in some details that I enjoy but that aren't super relevant to you, like what exactly to do with partial test results.

Perhaps the concrete bug you're seeing would help guide us to the right things to test. I guess you're running `update-bcd` with some results from Meta Quest, and it's making the wrong change in some cases? Can we start with one case to get warmed up?

unittest/unit/update-bcd.ts
```diff
@@ -996,7 +996,7 @@ describe('BCD updater', () => {
             actual,
             bcdFromSupport({
               chrome: {version_added: '85', flags: [{}]},
-              chrome_android: false
+              chrome_android: 'mirror'
```
jugglinmike, line 999: @foolip I've changed this expected value since you last reviewed, and I'd like to know if you agree with my reasoning.

The unit test describes the condition where the "upstream" browser support is "yes, but behind a flag", and the data says the feature is absent from the "downstream" browser.

To my mind, that data is consistent with the upstream browser, so the mirroring is still in effect, and the `"mirror"` value ought to be preserved.

What do you think?

unittest/unit/update-bcd.ts
jugglinmike: I can't collect test results from that browser at the moment; my participation in this repository has so far been proactive. I've been trying to enumerate all of the situations that the script might encounter when it receives the first set of results from the Quest browser. I may have identified some conditions which are more exotic than we'll see with authentic data, but I'm fairly confident that many represent mainline use cases. Any problems with those will block progress when I do have results in hand.

An example would be when a test result for a "downstream" browser is negative, and the "upstream" support is positive. The script ought to change `"mirror"` to `{version_added: false}`, but it currently does nothing. (That case is described by the second row of the table in gh-2261 and by the test named "BCD updater, update, mirror, supported upstream (without flags), unsupported in downstream test results" in this patch.)

It may be that we're closer than we think. Between [implementing support for the new "flags" data policy](https://github.com/foolip/mdn-bcd-collector/pull/2297), learning that existing support data should be ignored, and generally understanding the internals a bit better, I've been able to write a patch that corrects the problems. I'll be able to share it here once the "flags" patch is merged. In the mean time, [there's one more expected value I'd like to revisit](https://github.com/foolip/mdn-bcd-collector/pull/2280/commits/5f3e7f3ed6e7f299e18b04f946dd27be07a8bed7#r961206168).

foolip:
@jugglinmike do you want to fix the tests and un-draft this now?

unittest/unit/update-bcd.ts
foolip: Yes.

unittest/unit/update-bcd.ts
foolip: Yes, we should change `"mirror"` to `{version_added: false}` if applying the mirroring results in enabled-by-default support, which is contradicted by test results. This is the framework I'd suggest in general, to adjust data to be consistent with the evidence, and otherwise leave the data alone, even if we're not able to verify it.

jugglinmike:
@foolip I've corrected the tests and included a fix for the bug. Merging gh-2317 will almost certainly cause conflicts with this patch, and I'll be available to resolve those (they ought to be trivial, anyway).

update-bcd.ts
```diff
@@ -299,8 +299,9 @@ export const update = (
       continue;
     }
 
+    const support = entry.__compat.support;
     // Stringified then parsed to deep clone the support statements
-    const originalSupport = JSON.parse(JSON.stringify(entry.__compat.support));
+    const originalSupport = JSON.parse(JSON.stringify(support));
```
foolip, line 304: If you like, feel free to define `const clone = (value) => JSON.parse(JSON.stringify(value))` like in https://github.com/foolip/mdn-bcd-collector/pull/2317 and use it here/elsewhere.

foolip (approved):
Looks good in terms of the changes. Have you compared what the differences are in the results produced before/after this change? And do you want https://github.com/foolip/mdn-bcd-collector/pull/2317 or this landed first?

foolip:
@jugglinmike https://github.com/foolip/mdn-bcd-collector/pull/2317 is merged now. There are no conflicts, but do you want to add a `clone()` and show some concrete differences that result from the changes?

jugglinmike:
Sure! `clone` is now applied consistently throughout. As for the data, here's a summary of the changes 

I started with [this data from the Quest browser](https://github.com/foolip/mdn-bcd-results/compare/main...bocoup:mdn-bcd-results:quest). Using that, [here's what the "update" script on the `main` branch does to BCD](https://github.com/mdn/browser-compat-data/compare/main...bocoup:browser-compat-data:add-quest-from-unpatched-collector):

    $ git log -n2 --oneline add-quest-from-unpatched-collector
    cf127aea3 (add-quest-from-unpatched-collector) Add Quest data using unpatched BCD collector
    48a4a8d0a (upstream/main, upstream/HEAD, main) Remove "default values" notes for Web Audio constructors (#17811)
    $ git show --shortstat add-quest-from-unpatched-collector
    commit cf127aea34094bc821726d3a8b84af6b223f1c30 (add-quest-from-unpatched-collector)
    Author: Mike Pennisi <mike@mikepennisi.com>
    Date:   Tue Sep 20 13:06:58 2022 -0400
    
        Add Quest data using unpatched BCD collector
    
     9 files changed, 44 insertions(+), 15 deletions(-)

[And here's what the patched "update" script does to BCD](https://github.com/mdn/browser-compat-data/compare/main...bocoup:browser-compat-data:add-quest-from-patched-collector):

    $ git log -n2 --oneline add-quest-from-patched-collector
    713cd89ca (HEAD -> add-quest-from-patched-collector) Add Quest data using patched BCD collector
    48a4a8d0a (upstream/main, upstream/HEAD, main) Remove "default values" notes for Web Audio constructors (#17811)
    $ git show --shortstat add-quest-from-patched-collector
    commit 713cd89ca705bce7b365eb66abb467e454ae782c (HEAD -> add-quest-from-patched-collector)
    Author: Mike Pennisi <mike@mikepennisi.com>
    Date:   Tue Sep 20 13:09:57 2022 -0400
    
        Add Quest data using patched BCD collector
        
        See https://github.com/foolip/mdn-bcd-collector/pull/2280
    
     111 files changed, 1231 insertions(+), 404 deletions(-)

And here's a direct comparison of the two:

    $ git diff --shortstat add-quest-from-unpatched-collector add-quest-from-patched-collector
     103 files changed, 1187 insertions(+), 389 deletions(-)

There's too much to include inline, so [I've uploaded the full diff as a GitHub Gist](https://gist.github.com/jugglinmike/94e46ae4dab6635ed974ea40fe0985f9).

jugglinmike:
The required check for this job was cancelled. I don't know why that happened, and the output in CI doesn't help much:

    Checking formatting...
    All matched files use Prettier code style!
    Error: The operation was canceled.

foolip:
@jugglinmike in GitHub actions, when one job fails, the others are canceled. And it looks like the macOS job failed due to a timeout in `npm ci`. I'll just re-run the jobs, see if that was a network outage.

foolip:
Thanks @jugglinmike for the diff of the changes before/after these changes. Those looks sensible, but a number of them are probably due to problems with the tests like https://github.com/foolip/mdn-bcd-collector/issues/2347. But I have pretty high confidence in the changes based on that diff and the unit tests, of course.

foolip (approved):

jugglinmike:
Thank you, @foolip! I'll be scrutinizing the results over the coming days, and I'll keep you posted about any other anomalies I find.

foolip:
@jugglinmike thanks, looking forward to more bug reports! 😆

Issue #2282, opened 2022-08-30
add-new-bcd script cannot add CSS or JS features


foolip:
This script only considers the API category:

https://github.com/foolip/mdn-bcd-collector/blob/f8facc6cb7adebe28565f24d605ec8244eaffcae/add-new-bcd.ts#L126-L132

Fixing this bit of code isn't enough, however, there are other places that limit the scope to just API.

@queengooborg do you recall how this came about?

PR #2284, opened 2022-08-30
Support adding css.properties.* and javascript.builtins.* in add-new-bcd

foolip:
Fixes https://github.com/foolip/mdn-bcd-collector/issues/2282.


foolip:
@queengooborg feel free to review after the fact, I'll go ahead and merge now.

PR #2297, opened 2022-09-01
Remove flag data for default-enabled features

jugglinmike:
A recent change in the BCD project policy stipulates that information
about feature flags should be removed whenever a feature becomes
supported by default [1]. Modify the `update-bcd` script to honor the
new policy whenever new test results trigger that condition.

Include a test for the case where positive test results are added for a
feature that is already supported via a prefix. This patch intentionally
doesn't change that behavior because prefixes "are part of the web
platform that web developers have to deal with." [2] The test is to
prevent regressions and emphasize the intention.

[1] https://github.com/mdn/browser-compat-data/pull/16637
[2] https://matrix.to/#/!cGYYDEJwjktUBMSTQT:mozilla.org/$JTWeMpzNMULBTZcx-pWFnUNX7DIzE5xPySd70Kpr-MU?via=mozilla.org&via=matrix.org&via=igalia.com),

### DIFF
--- a/unittest/unit/update-bcd.ts
+++ b/unittest/unit/update-bcd.ts
@@ -68,6 +68,16 @@ const reports: Report[] = [
           result: true
         },
         {
+          name: 'api.UnflaggedInterface',
+          exposure: 'Window',
+          result: null
+        },
+        {
+          name: 'api.UnprefixedInterface',
+          exposure: 'Window',
+          result: null
+        },
+        {
           name: 'api.NullAPI',
           exposure: 'Window',
           result: null
@@ -148,6 +158,16 @@ const reports: Report[] = [
           result: true
         },
         {
+          name: 'api.UnflaggedInterface',
+          exposure: 'Window',
+          result: true
+        },
+        {
+          name: 'api.UnprefixedInterface',
+          exposure: 'Window',
+          result: true
+        },
+        {
           name: 'api.NewInterfaceNotInBCD',
           exposure: 'Window',
           result: false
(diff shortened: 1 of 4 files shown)


unittest/unit/update-bcd.ts
```diff
@@ -648,6 +710,8 @@ describe('BCD updater', () => {
         chrome: [{version_added: '0> ≤83', version_removed: '85'}]
       },
       'api.ExperimentalInterface': {chrome: [{version_added: '0> ≤83'}]},
+      'api.UnflaggedInterface': {chrome: [{version_added: '0> ≤84'}]},
```
foolip, line 713: Yeah, this `0>` is kind of weird :)

foolip (approved):

foolip:
Hmm, tests are actually failing?

jugglinmike:
My bad--didn't realize the test fixture was used in other contexts. It's fixed, now!

unittest/unit/update-bcd.ts
queengooborg: 0> should not be appearing in output values, hmm...

jugglinmike: GitHub's "review" interface is obscuring the context a bit, so it may not be clear that this particular test is for the `inferSupportStatements` function. That may include `0>` in its return value, but it may not be what you mean by "output," at least not from the perspective of someone using  the `update-bcd` script.

The `update` function, on the other hand, contains a handful of statements intended to remove `0>`. This patch exercises [this replacement](https://github.com/foolip/mdn-bcd-collector/pull/2297/files#diff-4cdb8f3a80ed4479be24afc5f4ec7f4c0f29c956731945d53a34691a97f0ede1R362-R365); disabling that code causes the test for the `update` function to fail (specifically for the `UnflaggedInterface` and `UnprefixedInterface` added here).

PR #2305, opened 2022-09-03
Set experimental to "true" for new features by default

queengooborg:
This PR updates the `add-new-bcd` script to set `experimental` to `true` by default.  If experimental should be false, the fix script will take care of it.


foolip (commented):
Thanks!

PR #2306, opened 2022-09-03
Remove duplicate IDL

queengooborg:
This PR removes the custom IDL that has been added to Webref, as well as fixes any issues related to it.


foolip:
I just added/changed both of these. Did you forget to `npm install` after the `@webref/IDL` update perhaps?

PR #2340, opened 2022-09-15
Support testing document.all

jugglinmike:
I validated this by running the collector locally and comparing the impact it has on the results. Only the results for `document.all` were affected, however, I have not generated a self-signed certificate, so this is only a limited view.


foolip (approved):
I think it might be possible to tweak `__test` instead, but this is fine :)

PR #2353, opened 2022-09-20
Formalize version range handling

jugglinmike:
Prior to this commit, the `version_added` property could take values which were undesirable for display to end-users (namely, those including the substring `0>`). Care was taken to avoid writing strings of this form to browser compatibility data structures.

Refactor the internals to preclude undesirable values, simplifying the code necessary to update the data structure and eliminating the risk of regression in future code paths.

### DIFF
--- a/update-bcd.ts
+++ b/update-bcd.ts
@@ -87,6 +87,21 @@ const combineResults = (results: TestResultValue[]): TestResultValue => {
   return supported;
 };
 
+// Create a string represenation of a version range, optimized for human
+// legibility.
+const joinRange = (lower: string, upper: string) =>
+  lower === '0' ? `≤${upper}` : `${lower}> ≤${upper}`;
+
+// Parse a version range string produced by `joinRange` into a lower and upper
+// boundary.
+const splitRange = (range: string) => {
+  const match = range.match(/(?:(.*)> )?(?:≤(.*))/);
+  if (!match) {
+    return {lower: range, upper: range};
+  }
+  return {lower: match[1] || '0', upper: match[2]};
+};
+
 // Get support map from BCD path to test result (null/true/false) for a single
 // report.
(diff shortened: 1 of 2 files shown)


jugglinmike:
See also https://github.com/foolip/mdn-bcd-collector/pull/2297#discussion_r961084002

update-bcd.ts
```diff
@@ -414,7 +425,7 @@ export const update = (
             simpleStatement.version_added !== 'preview' &&
             compareVersions(
               version,
-              simpleStatement.version_added.replace('≤', ''),
+              splitRange(simpleStatement.version_added).upper,
```
foolip, line 428: Is it really possible for `simpleStatement.version_added` to contain a `${lower}> ≤${upper}` range? Is this to handle the case where the update script is run twice?

This wasn't supported previously, and there aren't any tests that have this format in the initial data. If you want to support this, can you add a test?

update-bcd.ts
```diff
@@ -432,22 +443,14 @@ export const update = (
         typeof inferredStatement.version_added === 'string' &&
         inferredStatement.version_added.includes('≤')
       ) {
-        const range = inferredStatement.version_added.split('> ≤');
+        const {lower, upper} = splitRange(inferredStatement.version_added);
+        const simpleAdded = splitRange(simpleStatement.version_added).upper;
```
foolip, line 447: Ditto.

update-bcd.ts
jugglinmike: I don't want to introduce support for that condition, no. I thought using the utility function here (rather than a call to `replace`) would better communicate the code's intent. That's debatable, but it definitely makes the code unnecessarily forgiving. I've removed that application and replaced the code path designed to support it with a `throw` statement. Much safer!

foolip: Great, thanks!

foolip (approved):

PR #2365, opened 2022-09-23
Log warnings for when we do nothing in update-bcd

foolip:


foolip:
@jugglinmike FYI

jugglinmike:
Great idea!

PR #2371, opened 2022-09-23
Enable default (more verbose) mocha logger

foolip:


foolip:
I find it difficult to understand failures with this compact format.

PR #2400, opened 2022-10-05
Correct custom test for AudioProcessingEvent

jugglinmike:
This constructor requires two arguments, and the second must be an object whose properties conform to a specific interface:

https://webaudio.github.io/web-audio-api/#AudioProcessingEvent


foolip (approved):

foolip:
Thanks @jugglinmike! https://mdn-bcd-collector.appspot.com/tests/api/AudioProcessingEvent looks better now!

PR #2406, opened 2022-10-07
Add custom IDL for PermissionsPolicyViolationReportBody

foolip:
https://github.com/mdn/browser-compat-data/pull/17908#issuecomment-1271034331


foolip:
Leaving as draft since these tests won't pass, see https://github.com/mdn/browser-compat-data/pull/17908#issuecomment-1271322966.

PR #2568, opened 2022-12-26
Trying to add webassembly.features

MendyBerger:
Running `npm run update-bcd -- --path=webassembly.features*` doesn't create/modify the files in `browser-compat-data`

(I tried a few other things, but this didn't help either)


MendyBerger:
Mistake, never mind

PR #2582, opened 2023-01-03
Replace start-dev with dev to match Yari

foolip:
Cherry-picked from https://github.com/GooborgStudios/mdn-bcd-collector/commit/a9b3057a50d2e03f8d20d59c5b54e36144a72d08.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ These errors are worth looking out for:
 ## Running the server locally
 
 ```sh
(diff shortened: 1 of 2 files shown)


foolip:
Using "rebase and squash" resulted in the commit being attributed to me, so I force pushed a new commit to main. That loses which PR it came from, but whatever.

PR #2589, opened 2023-01-03
Fallback to legacy getUserMedia for MediaStream instance

foolip:
Cherry-picked from https://github.com/GooborgStudios/mdn-bcd-collector/commit/c27a1d7cccf348368fb13d6e5223705747e171ca.


foolip:
As in https://github.com/foolip/mdn-bcd-collector/pull/1991#issuecomment-1122475495 this breaks `MediaStreamTrack`, so the fix will need to be added here.

PR #2590, opened 2023-01-03
Migrate all scripts to TypeScript

foolip:
Cherry-picked from https://github.com/GooborgStudios/mdn-bcd-collector/commit/ff9cbcf4901954a9b1629c8fefccfec3fad9d03d.


foolip:
Nope, I can't get the production smoke test to work with simple means, so I'll disable it while I cherry-pick more changes.

Issue #2348, opened 2022-09-20
dataIsOlder early exit is too broad


foolip:
A `dataIsOlder` early exit in the update-bcd script was added in https://github.com/foolip/mdn-bcd-collector/pull/1904.

It looks like `dataIsOlder` is set to true if we have *any* data older than the BCD data, rather than all of the results being older. It was intended to fix https://github.com/foolip/mdn-bcd-collector/issues/659, which describes a problem where we have results from Safari 10-13, but a feature was added in Safari 14.

cc @queengooborg

foolip:
Fixed by https://github.com/foolip/mdn-bcd-collector/pull/2629.

PR #2629, opened 2023-01-04
Overhaul code to handle "false" but BCD set to newer version

foolip:
Fixes #15.

### DIFF
--- a/update-bcd.ts
+++ b/update-bcd.ts
@@ -462,38 +462,46 @@ export const update = (
         continue;
       }
 
-      let dataIsOlder = false;
+      // If we infer no support but BCD currently has a version number, check to make sure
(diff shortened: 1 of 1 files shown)


foolip:
This might fix https://github.com/foolip/mdn-bcd-collector/issues/2348.

update-bcd.ts
```diff
@@ -462,38 +462,46 @@ export const update = (
...
-            // A version we have data for is the same or newer than the version in BCD
-            dataIsOlder = true;
-            break;
+            latestNonNullVersion = version;
           }
         }
+
+        if (
+          compareVersions(
+            latestNonNullVersion,
```
foolip, line 492: Looks like it is safe to assume that `latestNonNullVersion` has been set to a version here, because if all of the results were null there should be no inferred support statements. This is tested by "api.NullAPI" in unit tests.

PR #2633, opened 2023-01-04
Replace the appspot.com URL wherever possible

foolip:
This is intended to reduce differences between forks.


foolip:
@queengooborg FYI

PR #2635, opened 2023-01-04
Move minimatch logic into update-bcd's update() function

foolip:
Pull out the handling to the start of the function rather than inside a
loop, and add types to make the separation between main() and update()
filter arguments a bit clearer. Categories are only used for loading the
data, and the filter is now always a string.


foolip:
@queengooborg FYI

PR #2650, opened 2023-01-05
Handle LegacyNoInterfaceObject

foolip:
The effect of this is to remove the tests for Metadata, which don't work
because this interface has [LegacyNoInterfaceObject] in Chromium.
Writing custom tests is hard because it requires accessing the local
file system and calling getMetadata() on an entry.

### DIFF
--- a/build.ts
+++ b/build.ts
@@ -777,6 +777,13 @@ const buildIDLTests = (ast, globals, scopes) => {
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


foolip:
This doesn't work because it also prevents custom tests from being generated. The only thing that this should do is to prevent the default prototype-based tests from being generated.

PR #2674, opened 2023-01-17
Fix some HTML spec violations

MrBrain295:
The stuff I removed was all outdated and unneeded.


foolip:
@queengooborg the `<meta http-equiv="Content-Type" content="text/html; charset=utf-8">` bit was added in https://github.com/foolip/mdn-bcd-collector/pull/416, do you know if it has an effect on any browser? My guess is no, but given how far back in browser history we've gone, I'm not sure.

MrBrain295:
I did some research and I don't think it will have any affect (except maybe on IE).

Issue #2576, opened 2023-01-02
Resolve differences to GooborgStudios version


foolip:
After https://github.com/foolip/mdn-bcd-collector/pull/2575#issuecomment-1368703465 I've realized that https://github.com/GooborgStudios/mdn-bcd-collector (deployed at https://mdn-bcd-collector.gooborg.com/) has diverged a fair bit from this repo, and there's no point in my continuing to update dependencies and deploy it unless we converge and have a single official deployment.

It looks like the last common commit is 06dcdc8b958aed77c2d5bbbfe327709f6a121689 from Nov 15.

The biggest differences are in custom IDL and custom tests defined by `custom-tests.yaml`.

@queengooborg are there are changes on my side you're aware of that need porting to your repo?

I think the most reasonable resolution to this is to archive my repo and redirect https://mdn-bcd-collector.appspot.com/ to https://mdn-bcd-collector.gooborg.com/, so that there's no ambiguity about who is maintaining the service actually used to update BCD.

Or, if you prefer, I can merge in all of your changes into this repo.

foolip:
I've been going through and porting over most changes, leaving ones (for now) that are about the deployment environment.

I still think we should have one official repo and deployment, but now think it's best to preserve the history (pull requests linked from BCD, etc) of this repo, but we could do that by transferring the repo.

In other words, I'd like to get them in sync by making changes to this repo.

queengooborg:
As I've mentioned a few times in other places, I will personally be continuing to maintain my own repo and utilize it for BCD updates, as previous deployment delays had severely impacted my ability to work on BCD.  Synchronizing does sound good, but I will not personally be spending time to submit pull requests or push commits to this repository, and focus my efforts primarily on BCD updates.

(Note: if you do pull changes over, I would like to request that the purple tint changes are left behind, as they are a) a part of Gooborg Studios branding, and b) a way I visually differentiate the two, since the domains are so similar.)

foolip:
[...]

PR #2714, opened 2023-03-15
Update tests

foolip:
- Add custom test for InkPresenter API
- Update tests for HTMLAllCollection to handle falsy property
- Update testConstructor with CompressionStream Safari constructor error


foolip:
Part of https://github.com/foolip/mdn-bcd-collector/issues/2576.

PR #2723, opened 2023-03-15
Filter out alt. names

foolip:
- Filter out alt. names for Document.characterSet
- Filter out alt. name for Window.navigator


foolip:
Filed https://github.com/foolip/mdn-bcd-collector/issues/2724 about a potential improvement.

PR #2758, opened 2023-03-23
[DNM] GCP Deployment

caugner:


app.yaml
```diff
@@ -8,4 +8,4 @@
 
 runtime: nodejs16
 env_variables:
-  GCLOUD_STORAGE_BUCKET: mdn-bcd-buffer
+  GCLOUD_STORAGE_BUCKET: moz-fx-dev-fiji-mdnplaygound.appspot.com
```
foolip, line 11: Is "playgound" a typo?

foolip:
@caugner let me know if/when you want review of this. Secrets management is something I've never been totally happy with, but I think the simplest way is to put secrets into GitHub Actions and create a `secrets.json` using them before deploy. You could also put secrets somewhere in GCP, but I don't know what the current best practices are.

foolip:
Also note https://github.com/foolip/mdn-bcd-collector/pull/2756. I'm not sure if you're trying to use the same nodejs version for all projects, if so feel free to change it to whatever you like.

foolip:
@caugner do you have an update on this and on the blockers for migrating the collector more generally? cc @Rumyra

foolip:
Ping @caugner for an update :)

caugner:
Closing, as I don't have any capacity to work on this in the near future.

PR #2824, opened 2023-09-26
Tombstone the foolip/mdn-bcd-collector repo

foolip:


foolip:
@queengooborg @Elchi3 FYI

